PR #94, opened 2014-08-31
deprecated the activation strategy class in favor of the one in Symfony

fabpot:
In symfony/symfony#11805, I propose to move the activation strategy class to the Symfont bridge. This PR deprecates the current class from the bundle in favor of the one in the bridge. Obviously, this PR can not be merged before the release of Symfony 2.6 at the earliest.


jum4:
Hi, 
What about this PR ?
'request' was removed in Symfony3 so NotFoundActivationStrategy does not work anymore.

Seldaek:
I guess at some point we need to branch out for symfony 3.0 (or 2.7+) into a v3.0 of this bundle.

Issue #37, opened 2013-03-28
Profiler enabled check in DebugHandlerPass doesn't work correctly


mac-cain13:
The [DebugHandlerPass](https://github.com/symfony/MonologBundle/blob/master/DependencyInjection/Compiler/DebugHandlerPass.php#L36-L38) checks if the profiler is enabled, but I think that this check is incomplete. I'm running a Symfony 2.2 instance and even in production mode the `profiler` section is defined in the config, only the `profiler.enabled` option is set to false to disable the profiler.

I think that the `DebugHandlerPass` class should check if the enabled is set to true and only then add the `DebugHandler`. Or is there something I'm missing?

PR #111, opened 2015-01-04
Add check to kernel.debug in DebugCompilerPass

yeroon:
When running CLI commands with the `--no-debug` option in the dev environment, the DebugHandler was still being registered. For long running commands with lots of logging (I like to use -vvv) this will cause memory issues.

The DebugCompilerPass now checks the `kernel.debug` parameter, suggested by @fabpot in #37.

### DIFF
--- a/DependencyInjection/Compiler/DebugHandlerPass.php
+++ b/DependencyInjection/Compiler/DebugHandlerPass.php
@@ -18,7 +18,7 @@ use Symfony\Component\DependencyInjection\Definition;
 use Monolog\Logger;
 
 /**
- * Adds the DebugHandler when the profiler is enabled or when kernel.debug is not false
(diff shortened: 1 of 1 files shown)


DependencyInjection/Compiler/DebugHandlerPass.php
```diff
@@ -18,7 +18,7 @@
 use Monolog\Logger;
 
 /**
- * Adds the DebugHandler when the profiler is enabled.
+ * Adds the DebugHandler when the profiler is enabled or when kernel.debug is not false.
```
xabbuh: Should be "and" not "or".

Seldaek:
Thanks

Seldaek:
Tagged 2.7.1 with the fix

PR #119, opened 2015-04-16
Added buffer handler config parameter

Matrooskin:


xabbuh:
You need to update the XML schema too.

Matrooskin:
Updated

DependencyInjection/Configuration.php
```diff
@@ -318,6 +319,7 @@ public function getConfigTreeBuilder()
                             ->scalarNode('min_level')->defaultValue('DEBUG')->end() // filter
                             ->scalarNode('max_level')->defaultValue('EMERGENCY')->end() //filter
                             ->scalarNode('buffer_size')->defaultValue(0)->end() // fingers_crossed and buffer
+                            ->scalarNode('flush_on_overflow')->defaultValue(false)->end() // buffer
```
stof: this should be a booleanNode (and please use the ``defaultFalse()`` method

Matrooskin: I fixed that line. Btw. https://github.com/symfony/MonologBundle/blob/master/DependencyInjection/Configuration.php#L302 :D

stof: the syslog ident is not a boolean

Seldaek:
Thanks

PR #120, opened 2015-05-05
Fix typos

maqnouch:


Seldaek:
That is actually introducing a typo so no thanks ;)

PR #121, opened 2015-05-12
separated wire of handlers to channels

nahpeps:
moved the wire process to a separate function to allow compiler passes
of other bundles to add custom log channels.

The additional function makes it more easy to add dynamic channels like this:

```php

namespace BaseBundle\DependencyInjection\CompilerPass;

use Symfony\Bundle\MonologBundle\DependencyInjection\Compiler\LoggerChannelPass;
use Symfony\Component\DependencyInjection\ContainerBuilder;
use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;

class LoggerChannelCompilerPass extends LoggerChannelPass {

    /**
     * @param ContainerBuilder $container
     * @api
     */
    {

        $serviceList = $container->getParameter('service_list');

        foreach ($serviceList as $serviceIdentifier) {
            $channel = 'my.channel_prefix.' .  $serviceIdentifier;
            $channelService = 'monolog.logger.my_channel_prefix.' . $serviceIdentifier;

            $this->createLogger($channel, $channelService, $container);
        }
        $this->wireHandlersToChannels($container);
    }
}

```

### DIFF
--- a/DependencyInjection/Compiler/LoggerChannelPass.php
+++ b/DependencyInjection/Compiler/LoggerChannelPass.php
@@ -69,18 +69,7 @@ class LoggerChannelPass implements CompilerPassInterface
         $container->getParameterBag()->remove('monolog.additional_channels');
 
         // wire handlers to channels
-        $handlersToChannels = $container->getParameter('monolog.handlers_to_channels');
-        foreach ($handlersToChannels as $handler => $channels) {
-            foreach ($this->processChannels($channels) as $channel) {
-                try {
-                    $logger = $container->getDefinition($channel === 'app' ? 'monolog.logger' : 'monolog.logger.'.$channel);
-                } catch (InvalidArgumentException $e) {
-                    $msg = 'Monolog configuration error: The logging channel "'.$channel.'" assigned to the "'.substr($handler, 16).'" handler does not exist.';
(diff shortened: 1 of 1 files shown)


stof:
what is the use case for that ? Why not just tagging the service with the ``monolog.logger`` tag ?

stof:
I'm not sure we want to expose ``wireHandlersToChannels`` as a supported extension point for which we need to keep BC.

nahpeps:
The use case for me is, that I do not want to have the logger injected by constructor but by a function call.
As far I know, it is not possible to define the channel by using using the "calls" setting in DI like this:

```yaml
s7_numbirds_importer.midoco2.import_controller:
        class: MyClass
        arguments: [...]
        calls:
           - [setLogger, ["@my_logger"]]
```

In my case, the services are started in another bundle and get the matching logger channel injected.

Seldaek:
@nahpeps I believe you can define it like that by injecting `@monolog.logger.foo` and then creating the channel in the config as documented at http://symfony.com/doc/current/cookbook/logging/channels_handlers.html#creating-your-own-channel

stof:
changing the channel when injecting the logger through a method call is supported since the 2.1.0 version of the bundle.
So this is a bad use case for this change.

PR #124, opened 2015-05-23
moving MonologBundle docs to bundle repo

snoek09:
Added installation chapter.

Easy picks on bundle side from symfony/symfony-docs#4983

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -17,8 +17,8 @@ in the `installation chapter`_ of the Composer documentation.
 Step 2: Enable the Bundle
 -------------------------
 
-Then, enable the bundle by adding the following line in the ``app/AppKernel.php``
-file of your project:
+Then, enable the bundle by adding it to the list of registered bundles
+in the ``app/AppKernel.php`` file of your project:
 
 .. code-block:: php
 
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -7,38 +7,39 @@ Step 1: Download the Bundle
 Open a command console, enter your project directory and execute the
 following command to download the latest stable version of this bundle:
 
-```bash
-$ composer require symfony/monolog-bundle "~2.7"
-```
+.. code-block:: bash
+
+    $ composer require symfony/monolog-bundle "~2.7"
 
 This command requires you to have Composer installed globally, as explained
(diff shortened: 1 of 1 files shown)


xabbuh:
looks good :+1:

WouterJ:
:+1:

Resources/doc/index.rst
```diff
@@ -0,0 +1,45 @@
...
+    $ composer require symfony/monolog-bundle "~2.7"
+
+This command requires you to have Composer installed globally, as explained
+in the `installation chapter`_ of the Composer documentation.
+
+Step 2: Enable the Bundle
+-------------------------
+
+Then, enable the bundle by adding the following line in the ``app/AppKernel.php``
+file of your project:
```
jakzal: > by adding the following line

Which line? The code example has more than a single line. 

Would be better to assume the user knows what to look for in the code example and say "Then enable the bundle by adding it to a list of registered bundles in the ``app/AppKernel.php`` file of your project:"

snoek09: That would be more clear. I would suggest changing this in the Installation Instructions template as well: http://symfony.com/doc/current/cookbook/bundles/best_practices.html#installation-instructions

xabbuh: see symfony/symfony-docs#5342

Resources/doc/index.rst
```diff
@@ -0,0 +1,45 @@
...
+
+Step 1: Download the Bundle
+---------------------------
+
+Open a command console, enter your project directory and execute the
+following command to download the latest stable version of this bundle:
+
+.. code-block:: bash
+
+    $ composer require symfony/monolog-bundle "~2.7"
```
aitboudad: I'd prefer to not specify the version only `composer require symfony/monolog-bundle` is enough.

WouterJ: It isn't, as it means the 2.7 docs would make people install 3.0 (if that version is released) as it would just take the latest stable version if you don't specify a version. This could lead to much confusion.

aitboudad: ahh ok ;)

stof:
is there anything left here ?

xabbuh:
Good to merge imho as @jakzal's comment has been addressed.

PR #126, opened 2015-06-08
Extend hipchat handler configuration to allow message_format config

rommsen:
This PR extends the hipchat handler configuration to allow the setting of different message_format (https://www.hipchat.com/docs/api/method/rooms/message) and the use of an ssl connection.

Both options are available in the constructor of the HipchatHandler (https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/HipChatHandler.php#L81) but were not configurable yet.

Please bear with me. This is my first Symfony related PR.


Seldaek:
Looks fine to me, thanks!

rommsen:
Thanks!

PR #129, opened 2015-06-09
update Travis CI configuration

xabbuh:
* use container based infrastructure
* add PHP 7 build job
* cache Composer cache directory on Travis
* report build status as fast as possible
* add build with lowest possible dependency versions
* run tests with different Symfony versions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,37 @@
+sudo: false
+
 language: php
+
+cache:
+    directories:
+        - $HOME/.composer/cache
 
 php:
   - 5.3
   - 5.4
   - 5.5
-  - 5.6
+  - 7.0
   - hhvm
 
-before_script:
+matrix:
+  fast_finish: true
+  include:
+    - 5.3
+      env: deps="low"
+    - php: 5.6
+      env: SYMFONY_VERSION="2.3.x"
+    - php: 5.6
+      env: SYMFONY_VERSION="2.6.x"
+    - php: 5.6
+      env: SYMFONY_VERSION="2.7.x"
+    - php: 5.6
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,14 +1,37 @@
...
+      env: SYMFONY_VERSION="2.3.x"
+    - php: 5.6
+      env: SYMFONY_VERSION="2.6.x"
+    - php: 5.6
+      env: SYMFONY_VERSION="2.7.x"
+    - php: 5.6
+      env: SYMFONY_VERSION="2.8.x-dev"
+
+before_install:
+    - if [ "$SYMFONY_VERSION" != "" ]; then composer require --dev --no-update symfony/symfony:"$SYMFONY_VERSION"; fi
```
stof: you should do that after updating composer itself (or not update composer)

.travis.yml
```diff
@@ -1,14 +1,37 @@
...
     - composer self-update
-    - composer install --prefer-source
+
+install:
+    - if [ "$deps" = "low" ]; composer update --prefer-lowest; else composer install --prefer-source; fi
```
stof: ``--prefer-source`` should be removed, as it kills all benefits of persisting the cache (downloaded archives go in the cache, but not source installs)

.travis.yml
Seldaek: Seems fine to me as it's done with a --no-update

.travis.yml
Seldaek: Missed that, I'll fix directly

xabbuh: Oh indeed, thanks for catching this @stof

PR #130, opened 2015-06-22
Fixed travis-ci

zerkms:
At the moment `.travis.yml` [is broken](http://lint.travis-ci.org/symfony/MonologBundle):

1. `php:` key is missing
2. `then` is missing in a bash command


stof:
Thank you @zerkms.

Issue #122, opened 2015-05-13
Parameterized channel


zerkms:
At the moment the channel parameter is taken as-is:

                $loggerId = sprintf('monolog.logger.%s', $tag['channel']);
                $this->createLogger($tag['channel'], $loggerId, $container);

whereas I find it useful if it was possible to make it a parameter so that you only needed to specify/change it in one place.

Thoughts?

Seldaek:
Sorry but not sure what you mean.. Do you mean you'd like if it resolved the `$tag['channel']` for parameters before using it? If so that sounds alright to me and please send a pull request :)

zerkms:
Good :-)

And, yes, that's what I meant

Taluu:
@zerkms you should maybe close this ticket as the resulting PR was merged ;)

PR #131, opened 2015-06-22
Added resolving channel name via container parameters

zerkms:
It's a PR for #122

Now it's possible to specify channel as a parameter in a tagged configuration:

XML:

    <services>
        <service id="foo_logger" class="Foo">
            <tag name="monolog.logger" channel="%channel_parameter%" />
        </service>
    </services>

YML:

    services:
        foo_logger:
            class: Foo
            tags: [{ name: monolog.logger, channel: %channel_parameter% }]


Seldaek:
Thanks, looks good!

PR #134, opened 2015-08-18
Adding a Swift MessageFactory for performance

weaverryan:
Hiya guys!

Blackfire comparison: https://blackfire.io/profiles/compare/0ab8c333-3272-4af0-8b57-ff0b4e3478de/graph

Prior to this, in order for the main logger to be instantiated, if you had a swiftmailer handler, ultimately a single Swift_Message would be created, even if you didn't need it. That sounds minor, but due to Swift's
autoloading, it required quite a few extra classes, which added a small amount of extra memory (~1MB) and extra CPU due to the auto-loader (~10-20ms).

This keeps BC, only renaming a service that was private anyways.

Cheers!

### DIFF
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -390,33 +390,21 @@ class MonologExtension extends Extension
                     $prototype = new Reference($handler['email_prototype']['id']);
                 }
             } else {
-                $message = new Definition('Swift_Message');
-                $message->setLazy($handler['lazy']);
-                $message->setPublic(false);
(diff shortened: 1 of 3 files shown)


Seldaek:
Seems good to me, @stof any feedback?

weaverryan:
Rebased and tests passing. I think this should be safe - it only changes a non-public service, and is doing so on the master branch anyways.

Thanks!

Seldaek:
Alright then, thanks

weaverryan:
@Seldaek aw shucks, you're the best :)

PR #138, opened 2015-09-15
Reference::isStrict() is gone in Symfony 3.0

derrabus:
The `LoggerChannelPass` contains calls to `Symfony\Component\DependencyInjection\Reference::isStrict()`, a method that has been removed from Symfony 3. The test case `LoggerChannelPassTest` fails if you run it against the Symfony master branch.

This pull request adds a forward-compatible code that checks with `method_exists()` if the method is still present. That check may be removed in the future, as soon as support for Symfony 2 is dropped. Tests are now running green with Symfony 2.7.4 and 3.0-dev.

### DIFF
--- a/DependencyInjection/Compiler/LoggerChannelPass.php
+++ b/DependencyInjection/Compiler/LoggerChannelPass.php
@@ -85,24 +85,6 @@ class LoggerChannelPass implements CompilerPassInterface
         }
     }
 
-    /**
-     * Creates a copy of a reference and alters the service ID.
-     *
-     * @param Reference $reference
-     * @param string    $serviceId
-     *
-     * @return Reference
-     */
-    private function changeReference(Reference $reference, $serviceId)
-    {
-        if (method_exists($reference, 'isStrict')) {
-            // Stay compatible with Symfony 2
-            return new Reference($serviceId, $reference->getInvalidBehavior(), $reference->isStrict());
-        } else {
-            return new Reference($serviceId, $reference->getInvalidBehavior());
-        }
-    }
-
     public function getChannels()
     {
         return $this->channels;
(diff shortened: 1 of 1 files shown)


derrabus:
The method was removed with PR symfony/symfony#15693.

fabpot:
:+1:

fabpot:
That's actually the only remaining patch to make the Symfony SE work for Symfony 3.0. At least, to make Composer happy :)

DependencyInjection/Compiler/LoggerChannelPass.php
```diff
@@ -85,6 +85,24 @@ public function process(ContainerBuilder $container)
...
+     * @param string    $serviceId
+     *
+     * @return Reference
+     */
+    private function changeReference(Reference $reference, $serviceId)
+    {
+        if (method_exists($reference, 'isStrict')) {
+            // Stay compatible with Symfony 2
+            return new Reference($serviceId, $reference->getInvalidBehavior(), $reference->isStrict());
+        } else {
```
stof: no need to use ``else`` here as the previous ``if`` returns

DependencyInjection/Compiler/LoggerChannelPass.php
```diff
@@ -85,6 +85,24 @@ public function process(ContainerBuilder $container)
...
 
+    /**
+     * Creates a copy of a reference and alters the service ID.
+     *
+     * @param Reference $reference
+     * @param string    $serviceId
+     *
+     * @return Reference
+     */
+    private function changeReference(Reference $reference, $serviceId)
```
stof: the private method should be defined after public and protected ones

derrabus:
@stof Thank you for your review. I've followed both of your suggestions.

stof:
Thank you @derrabus.

DependencyInjection/Compiler/LoggerChannelPass.php
```diff
@@ -112,4 +112,22 @@ protected function createLogger($channel, $loggerId, ContainerBuilder $container
...
+    {
+        if (method_exists($reference, 'isStrict')) {
+            // Stay compatible with Symfony 2
+            return new Reference($serviceId, $reference->getInvalidBehavior(), $reference->isStrict());
```
Tobion: This still triggers alot of deprecation warnings when warming the cache in the standard edition. We need a different solution. We could just use `$reference->isStrict(false)` to prevent the deprecation warning. Or can we not ignore the `strict` flag completely? It's rather unlikely to have a scope problem with a logger...

derrabus: `$reference->isStrict(false)` is probably the way to go. Ignoring the strict flag might cause problems with bundles that still rely on the deprecated behavior.

PR #139, opened 2015-09-20
Default swiftmailer transport was always used

### LINKED ISSUE #137: Wrong transport used in AddSwiftMailerTransportPass
If we are using several mailers configurations with swiftmailer:
```yaml
# app/config/config.yml
swiftmailer:
    default_mailer: default
    mailers:
        default:
            ...
        errors:
            ...
```

and specifying a non-default mailer in monolog:

```yaml
# app/config/config_prod.yml
monolog:
    handlers:
        mail:
            type:         fingers_crossed
            action_level: critical
            handler:      buffered
        buffered:
            type:    buffer
            handler: swift
        swift:
            type:       swift_mailer
            # This is the specified mailer service :
            mailer:     swiftmailer.mailer.errors 
            from_email: error@example.com
            to_email:   error@example.com
            level:      debug
```

When doing this, the definition of the monolog handler is correct. 
But the problem is in the class AddSwiftMailerTransportPass of MonologBundle ([current version](https://github.com/symfony/MonologBundle/blob/7b3658a301a67f8776026e629d53e8e75c0f582e/DependencyInjection/Compiler/AddSwiftMailerTransportPass.php)), this CompilerPass adds the `setTransport` method to the handler definition with the default transport. ([Here](https://github.com/symfony/MonologBundle/blob/7b3658a301a67f8776026e629d53e8e75c0f582e/DependencyInjection/Compiler/AddSwiftMailerTransportPass.php#L42) and [here](https://github.com/symfony/MonologBundle/blob/7b3658a301a67f8776026e629d53e8e75c0f582e/DependencyInjection/Compiler/AddSwiftMailerTransportPass.php#L50))

When using only one mailer, it's working.

When using a non-default mailer in monolog, the mails are not sent because the transport (aka the spool) used is the one of the default mailer.

I made it work but it's not perfect :
```php
<?php

namespace Symfony\Bundle\MonologBundle\DependencyInjection\Compiler;

use Symfony\Component\DependencyInjection\ContainerBuilder;
use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;

class AddSwiftMailerTransportPass implements CompilerPassInterface
{
    {
        $handlers = $container->getParameter('monolog.swift_mailer.handlers');

        foreach ($handlers as $id) {
            $definition = $container->getDefinition($id);

            // This is my current solution
            // I'm retrieving the argument n°0 of this monolog handler definition 
            // (arg n°0 is a reference to the swiftmailer mailer), so when
            // calling _toString(), I can retrieve the id of the mailer and use
            // it for building the id of the correct transport.
            $mailerId = $definition->getArgument(0)->__toString();

            if (
                $container->hasAlias($mailerId . '.transport.real') ||
                $container->hasDefinition($mailerId . '.transport.real')
            ) {
                $definition->addMethodCall(
                     'setTransport',
                     array(new Reference($mailerId . '.transport.real'))
                 );
            } elseif (
                $container->hasAlias($mailerId . '.transport') ||
                $container->hasDefinition($mailerId . '.transport')
            ) {
                $definition->addMethodCall(
                     'setTransport',
                     array(new Reference($mailerId . '.transport'))
                 );
            }
        }
    }
}
```

What do you think of this solution ?

Does anybody has a better way to fix this bug ?

tseho:
The transport of the default mailer was always used, even if the monolog handler
has a different mailer setting. When using a non-default mailer, the correct transport
was never called and the monolog emails never sent.

Instead of assuming the default mailer transport, I now set the transport of the used
mailer.

Fixes #137.

### DIFF
--- a/DependencyInjection/Compiler/AddSwiftMailerTransportPass.php
+++ b/DependencyInjection/Compiler/AddSwiftMailerTransportPass.php
@@ -33,21 +33,23 @@ class AddSwiftMailerTransportPass implements CompilerPassInterface
         foreach ($handlers as $id) {
             $definition = $container->getDefinition($id);
 
+            $mailerId = $definition->getArgument(0)->_toString();
+
             if (
-                $container->hasAlias('swiftmailer.transport.real') ||
-                $container->hasDefinition('swiftmailer.transport.real')
+                $container->hasAlias($mailerId . '.transport.real') ||
+                $container->hasDefinition($mailerId . '.transport.real')
             ) {
                 $definition->addMethodCall(
                     'setTransport',
-                    array(new Reference('swiftmailer.transport.real'))
+                    array(new Reference($mailerId . '.transport.real'))
                 );
             } elseif (
-                $container->hasAlias('swiftmailer.transport') ||
-                $container->hasDefinition('swiftmailer.transport')
+                $container->hasAlias($mailerId . '.transport') ||
+                $container->hasDefinition($mailerId . '.transport')
             ) {
                 $definition->addMethodCall(
                     'setTransport',
-                    array(new Reference('swiftmailer.transport'))
+                    array(new Reference($mailerId . '.transport'))
                 );
             }
         }


Seldaek:
Thanks, I think it makes sense, I hope there is no edge case where this fails, but AFAIK after a quick look at the Swiftmailer DI Extension it seems fine.

PR #140, opened 2015-10-02
Fixed BC break in AdDSwiftMailerTransportPass

Nyholm:
This [commit](https://github.com/symfony/monolog-bundle/commit/159129fd3d15e37c49756f71c3204b56860026bc) introduced a BC break. 

The suggested configuration from the [Symfony docs](http://symfony.com/doc/current/cookbook/logging/monolog_email.html) look like this: 

```yaml
# app/config/config_prod.yml
monolog:
    handlers:
        mail:
            type:         fingers_crossed
            action_level: critical
            handler:      buffered
        buffered:
            type:    buffer
            handler: swift
        swift:
            type:       swift_mailer
            from_email: error@example.com
            to_email:   error@example.com
            level:      debug
```
The value of the handers['swift']['mailer'] will be default to 'mailer'. This is okey because the SwiftBundle has an alias for 'mailer'. (`$container->setAlias('mailer', 'swiftmailer.mailer');`).

When we use the 'mailer' in our `AddSwiftMailerTransportPass` we are looking for a service named `mailer.transport.real´ which does not exist and we will face an error like this: 
```
Uncaught exception 'Exception' with message 'No transport available to flush mail queue' in xx/vendor/symfony/symfony/src/Symfony/Bridge/Monolog/Handler/SwiftMailerHandler.php:85Stack trace:
#0 xx/vendor/symfony/symfony/src/Symfony/Bridge/Monolog/Handler/SwiftMailerHandler.php(65): Symfony\Bridge\Monolog\Handler\SwiftMailerHandler->flushMemorySpool()
#1 xx/vendor/monolog/monolog/src/Monolog/Handler/MailHandler.php(36): Symfony\Bridge\Monolog\Handler\SwiftMailerHandler->send('[2015-10-02 11:...', Array)
#2 xx/vendor/monolog/monolog/src/Monolog/Handler/BufferHandler.php(90): Monolog\Handler\MailHandler->handleBatch(Array)
#3 xx/vendor/monolog/monolog/src/Monolog/Handler/BufferHandler.php(106): Monolog\Handler\BufferHandler->flush()#4 [internal function]: Monolog\Handler\BufferHandler->close()#5 {main} thrown
```

This PR make sure we don't break BC and fall back on the `'swiftmailer.transport.real'` if no configured transport exists

### DIFF
--- a/DependencyInjection/Compiler/AddSwiftMailerTransportPass.php
+++ b/DependencyInjection/Compiler/AddSwiftMailerTransportPass.php
@@ -24,7 +24,7 @@ use Symfony\Component\DependencyInjection\Reference;
 class AddSwiftMailerTransportPass implements CompilerPassInterface
 {
     /**
-     * {@inheritDoc}
+     * {@inheritdoc}
      */
     public function process(ContainerBuilder $container)
     {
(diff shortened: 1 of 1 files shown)


Nyholm:
Ping @tseho, I would like to have your opinion on this. Have I understood you commit correctly? What do you think about my change?

tseho:
@Nyholm You are right, I thought I tested it correctly.

We DO need to fallback on "swiftmailer.transport" instead of "mailer.transport".

ping @Seldaek for approval.

Seldaek:
Seems legit. Thanks

Nyholm:
Thank you both for quick response!

PR #141, opened 2015-10-29
Don't trigger a deprecation warning for Reference::isStrict()

derrabus:
As @Tobion pointed out in a comment to my merged PR #138, the `LoggerChannelPass` still triggers deprecation warnings with Symfony 2.8. When writing my patch, I was not aware that I could suppress the warning by passing `false` as parameter to `Reference::isStrict()`.

I think, suppressing the warning is the way to go because we already took care of the deprecation in #138: Once `isStrict()` is gone, it won't be called anymore. But as long as it's still there, other bundles might be using that functionality, so imho there's no way around calling `isStrict()` without potentially breaking stuff.


nicolas-grekas:
:+1:

Tobion:
:+1:

Seldaek:
Thanks

PR #142, opened 2015-10-30
Allow authentication against protected Elasticsearch server

damien-list:
Hi,

Just add the possibility to authenticate against a protected Elasticsearch server.
Also allow to connect to Elasticsearch server behind SSL.

First pull request here, let me know if something's wrong ;)


xabbuh:
You have to update the XSD file too.

damien-list:
Here we go :)

Seldaek:
Thanks!

PR #143, opened 2015-11-04
Allow to autowire the default logger

dunglas:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | n/a
| License       | MIT
| Doc PR        | n/a

Related to symfony/symfony#16464. cc @weaverryan

### DIFF
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -53,6 +53,11 @@ class MonologExtension extends Extension
             $loader->load('monolog.xml');
             $container->setAlias('logger', 'monolog.logger');
 
+            // always autowire the main logger, require Symfony >= 2.8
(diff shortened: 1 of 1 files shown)


weaverryan:
:+1: 

The idea is that when a bundle introduces multiple services with the same interface/class, that service cannot be auto-wired as a dependency into user-land services, as the container doesn't know which service to auto-wire. By adding this, the main logger will be auto-wired, which is really what you want. The user can still overwrite this in their code by creating a new service with this interface and setting its auto-write type.

@dunglas I assume that the `monolog.logger` tag would still change the auto-wired `@logger` service? http://symfony.com/doc/current/reference/dic_tags.html#monolog-logger

dunglas:
Yes I've tried it it works.

Seldaek:
Thanks!

PR #146, opened 2015-11-12
Expose all of HipChat's parameters

apfelbox:
There are two currently unexposed parameters of the [`HipChatHandler`](https://github.com/Seldaek/monolog/blob/f86e643a5814f90724ddd09b99304611633f92e0/src/Monolog/Handler/HipChatHandler.php#L98): 

* `$host` which is required to support [HipChat Server](https://www.hipchat.com/server)
* `$version` the API version, where you might want to switch to the v2 api

These two options are currently not exposed in the configuration.

I wasn't sure on how to add these options and especially unsure about how to define a fallback in the configuration, if the option is used by multiple handlers (like `host`). The current way is to just pass `null` in these parameters and set the defaults in the handler directly:

```php
public function __construct($token, $room, $name = 'Monolog', $notify = false, $level = Logger::CRITICAL, $bubble = true, $useSSL = true, $format = 'text', $host = 'api.hipchat.com', $version = self::API_V1)
{
    if (null === $host)
    {
        $host = 'api.hipchat.com';
    }

    if (null === $version)
    {
        $version = self::API_V1;
    }

    // ...
}
```

Please tell me if whether this is ok or how I can implement it in a better way.
If the proposed implementation is ok, I would send a PR to Monolog for the changes mentioned above.

This would however be a soft BC break (in the sense that the `monolog-bundle` then requires the newest version of `monolog`).

### DIFF
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -468,6 +468,8 @@ class MonologExtension extends Extension
                 $handler['bubble'],
                 $handler['use_ssl'],
                 $handler['message_format'],
+                !empty($handler['host']) ? $handler['host'] : 'api.hipchat.com',
+                !empty($handler['api_version']) ? $handler['api_version'] : 'v1',
             ));
             break;
 
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -179,6 +179,8 @@ use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
  *   - [bubble]: bool, defaults to true
  *   - [use_ssl]: bool, defaults to true
  *   - [message_format]: text or html, defaults to text
+ *   - [host]: defaults to "api.hipchat.com"
+ *   - [api_version]: defaults to "v1"
  *
  * - slack:
  *   - token: slack api token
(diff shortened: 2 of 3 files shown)


apfelbox:
This is especially important, as there currently is now way to switch the API via the configuration and the `v1` api (the default one) is deprecated (https://www.hipchat.com/docs/api).

DependencyInjection/MonologExtension.php
```diff
@@ -468,6 +468,8 @@ private function buildHandler(ContainerBuilder $container, $name, array $handler
                 $handler['bubble'],
                 $handler['use_ssl'],
                 $handler['message_format'],
+                $handler['host'],
```
Seldaek: I tihnk you can just pass 'api.hipchat.com' as value for the host here in case it's null, that would solve the need for a change in monolog

apfelbox: Ok, I wasn't sure about that. If the php rfc for `default` had passed, I would pass that here (to keep the default value in monolog), but I can change it manually to the default value.

Should I do the same for the `api_version` parameter?

Seldaek: Oh sure doing the same there makes sense, didn't realize it was needed.

stof: @apfelbox you could still not pass default here. You would have to pass a special value supported by Symfony to dump it as the keyword ``default``

Seldaek: @stof this sort of comment is really not helpful IMO, and I don't need to get fact-checking of hypothetical statements in my inbox :/

apfelbox:
@stof true, but I think symfony would have built in a logic to automatically extract the default value in case of `default` when dumping. :wink: 

@Seldaek updated. I extracted the default values of both parameters into this bundle.

* I used `empty` because other similar checks in the file use it, too.
* No BC breaks
* I didn't add the default value for `api_version` directly in the configuration class (`->defaultValue("v1")`), as `api_version` looks like a value that might be used for multiple handlers in the future and they might name their API versions differently.

Seldaek:
Thanks!

PR #149, opened 2015-11-19
Default elasticsearch handler transport uppercase first character

damour:
With default elasticksearch handler transport I have exception on linux machine: 'Invalid transport'.
In elastica client liblary default transport value is 'Http' not 'http'.
https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Connection.php#L29
https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Transport/AbstractTransport.php#L92


Seldaek:
Thanks, I also sent a PR to elastica because this is not cool ;)

PR #151, opened 2015-12-22
Improve Documentation

d3f3kt:
Add a small notice to the slack configuration that the channel name have to contain the '#'.


Seldaek:
Thanks

PR #153, opened 2016-01-13
Add include_stacktraces configuration option to handlers

nicolas-grekas:
Fixes https://github.com/symfony/symfony/pull/17168
Allows people to add stacktraces to their log when they want to.


iltar:
:+1: 

If you can add a bit of documentation how to set this up to only work for certain log levels, that would be really awesome! I'm not interested in seeing warnings, info or even debug. I only want to see the important stuff such as errors or criticals for example.

PR #157, opened 2016-02-10
Add a nested option on handler config definition

wpottier:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Marked nested handlers with not be pushed to the logger handler array.
This options may be usefull when using custom handler (as service type handlers) which rely on a sub handler defined in the monolog-bundle configuration.
Marking this subhandler as nested avoid monolog to directly call it.

### Sample Usage

```yaml
monolog:
    handlers:
        foo:
            type: service
            id: my_site.custom_log_handler
        bar:
            type:  stream
            path:  "%kernel.logs_dir%/%kernel.environment%-bar.log"
            level: debug
            nested: true

services:
    my_site.custom_log_handler:
        class: AppBundle\Logger\CustomHandler
        arguments:
            - "@monolog.handler.bar"
```

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -592,6 +592,7 @@ class Configuration implements ConfigurationInterface
                                 ->end()
                             ->end()
                             ->scalarNode('formatter')->end()
+                            ->booleanNode('nested')->defaultFalse()->end()
                         ->end()
                         ->validate()
                             ->ifTrue(function ($v) { return 'service' === $v['type'] && !empty($v['formatter']); })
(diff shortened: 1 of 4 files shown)


wpottier:
May i've missed something or this PR didn't interested you?

Seldaek:
No it looks good to me, I just missed it :)

wpottier:
no problem. thanks a lot for your work !

PR #159, opened 2016-02-24
Add timeout and conection timeout to logentries handler config definition

dcsg:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        |

### DIFF
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -572,6 +572,12 @@ class MonologExtension extends Extension
                 $handler['level'],
                 $handler['bubble'],
             ));
+            if (isset($handler['timeout'])) {
+                $definition->addMethodCall('setTimeout', array($handler['timeout']));
+            }
+            if (isset($handler['connection_timeout'])) {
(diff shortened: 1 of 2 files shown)


xabbuh:
You need to add the new options to the XSD file too.

DependencyInjection/MonologExtension.php
```diff
@@ -572,6 +572,12 @@ private function buildHandler(ContainerBuilder $container, $name, array $handler
                 $handler['level'],
                 $handler['bubble'],
             ));
+            if (isset($handler['timeout'])) {
+                $definition->addMethodCall('setTimeout', [$handler['timeout']]);
```
xabbuh: The short array notation is not compatible with PHP 5.3.

dcsg:
@xabbuh they are not "new" since they are already defined for socket handler. But yes I don't see any timeout or connection_timeout in the .xsd file

Seldaek:
Thanks, I added the missing stuff to xsd

PR #160, opened 2016-02-26
Add `auto_log_stacks` param support for Sentry (Raven Client)

AkshayBz:
This will also serve as future reference for adding additional Raven Client params.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -165,6 +165,7 @@ use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
  *   - client_id: Raven client custom service id (optional)
  *   - [level]: level name or int value, defaults to DEBUG
  *   - [bubble]: bool, defaults to true
+ *   - [auto_stack_logs]: bool, defaults to false
  *
  * - newrelic:
  *   - [level]: level name or int value, defaults to DEBUG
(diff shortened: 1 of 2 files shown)


dcramer:
This is probably a good default for the Sentry integration.

The primary issue here is also present in Laravel: exceptions aren't bubbled up appropriate to Sentry, therefore it can't create a reliable backtrace. Monolog (I think?) seems to stuff these into context, but that's not overly useful as its unstructured.

Ideally Sentry itself would just hook into Symfony (and Laravel) native error handlers, vs using a logging abstraction, but that may not be easily doable.

AkshayBz:
@dcramer - That's true.
screen shot 2016-02-27 at 3 42 10 am
As you can see in the screenshot, out of the 12 entries in stacktrace, the top 4 are part of the Monolog code trace
PS: These additional entries are not seen if provide exception in the context.

AkshayBz:
@stof - Please have a look

Seldaek:
PR looks good to me but I have no knowledge about Sentry unfortunately. 

> Monolog (I think?) seems to stuff these into context, but that's not overly useful as its unstructured.

This is true, `$msg['context']['exception']` is the recommended way to store Exception instances, as per PSR-3 so it is kinda structured or at least standardized.

If you two agree that this should be merged then I guess I'll do it. Let me know.

AkshayBz:
Please do merge the PR.
`auto_log_stacks` is definitely a useful option to have in Sentry.
I'll try a few things to improve Monolog's Sentry support. Will create another PR if the experiments succeed.

PR #168, opened 2016-05-04
Removed unused file

### LINKED ISSUE #167: `NotFoundActivationStrategy.php` file should be removed
Since it's not used by the bundle and only confuses

zerkms:
NotFoundActivationStrategy class is shipped by Symfony

Closes #167


xabbuh:
But removing the file is a BC break for people who use this class.

zerkms:
@xabbuh fair point, but not obvious why would someone use it directly

zerkms:
May be mark it as deprecated then?

PR #170, opened 2016-07-12
remove class parameters

avant1:
The idea was initially discussed in symfony/symfony/issues/11881.

If I understand things correctly, we should provide a service for each removed parameter to allow bundle users to decorate or override corresponding services.
But I'm not sure if services should be listed in DI config of bundle, or it is OK to add them during DI Extension load.

I'm using those services to map handler type to handler class (this is done [here](https://github.com/avant1/monolog-bundle/blob/remove-class-parameters/DependencyInjection/MonologExtension.php#L133-L134)). But those handler services can be replaced by hash mapping.

If anything else related to this changes should be done (like updating README or adding note to symfony logging cookbook) - I'm ready to work on this too.

### DIFF
--- a/Resources/config/monolog.xml
+++ b/Resources/config/monolog.xml
@@ -4,53 +4,6 @@
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
 
-    <parameters>
-        <parameter key="monolog.logger.class">Symfony\Bridge\Monolog\Logger</parameter>
-        <parameter key="monolog.gelf.publisher.class">Gelf\MessagePublisher</parameter>
-        <parameter key="monolog.gelfphp.publisher.class">Gelf\Publisher</parameter>
-        <parameter key="monolog.handler.stream.class">Monolog\Handler\StreamHandler</parameter>
-        <parameter key="monolog.handler.console.class">Symfony\Bridge\Monolog\Handler\ConsoleHandler</parameter>
-        <parameter key="monolog.handler.group.class">Monolog\Handler\GroupHandler</parameter>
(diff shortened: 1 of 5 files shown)


Resources/config/monolog.xml
```diff
@@ -61,10 +14,53 @@
...
+        <service id="monolog.handler.whatfailuregroup" class="Monolog\Handler\WhatFailureGroupHandler"/>
+        <service id="monolog.activation_strategy.not_found" class="Symfony\Bundle\MonologBundle\NotFoundActivationStrategy"/>
+
+        <service id="monolog.handler.fingers_crossed" class="Monolog\Handler\FingersCrossedHandler"/>
+        <service id="monolog.handler.fingers_crossed.error_level_activation_strategy" class="Monolog\Handler\FingersCrossed\ErrorLevelActivationStrategy"/>
+        <service id="monolog.handler.filter" class="Monolog\Handler\FilterHandler"/>
+        <service id="monolog.handler.mongo" class="Monolog\Handler\MongoDBHandler"/>
+        <service id="monolog.mongo.client" class="MongoClient"/>
+
+        <service id="monolog.handler.elasticsearch" class="Monolog\Handler\ElasticSearchHandler"/>
```
xabbuh: The service definitions should be removed here (they will be registered by the extension only if needed).

avant1: @xabbuh indeed, now I understand that handlers services should not be listed here.
I've updated PR.

javiereguiluz:
👍 this PR looks nice and it complies with the modern Symfony practices. I hope that bundle maintainers will consider to merge it. Thanks!

stof:
this should be merged only when we decide to make a 2.0 version of the bundle, as it is a BC break

fabpot:
Thank you @avant1.

xabbuh:
@fabpot It looks like the `v2.11.2` tag was created on the wrong branch.

fabpot:
Sorry about that. Just released 2.11.3.

PR #171, opened 2016-07-27
JsonFormatter does support includeStacktraces

Nicofuma:


stof:
Good catch, thanks @Nicofuma.

PR #172, opened 2016-08-26
Added content-type to schema

jgroth:
content-type is missing from the XSD. making it impossible to use the swift_mailer handler to send HTML formatted emails with XML config


Seldaek:
Thanks

PR #173, opened 2016-09-15
ability to configure whether messages should be formatted as PSR-3

ged15:
This PR https://github.com/symfony/symfony/pull/17166 (reasons described in https://github.com/symfony/symfony/issues/15753) has changed the way messages are logged in Symfony. Now log messages contain placeholders (see https://github.com/symfony/symfony-standard/issues/981).
While this could be a useful feature, IMO by default in SE messages should be preformatted.
This PR introduces a new optional config option that makes the messages be preformatted by default but this behaviour can also be disabled.

### DIFF
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -138,6 +138,17 @@ class MonologExtension extends Extension
             $definition->setConfigurator(array('Symfony\\Bundle\\MonologBundle\\MonologBundle', 'includeStacktraces'));
         }
 
+        if ($handler['process_psr_3_messages']) {
+            $processorId = 'monolog.processor.psr_log_message';
+            if (!$container->hasDefinition($processorId)) {
+                $processor = new Definition('Monolog\\Processor\\PsrLogMessageProcessor');
+                $processor->setPublic(false);
+                $container->setDefinition($processorId, $processor);
+            }
+
+            $definition->addMethodCall('pushProcessor', array(new Reference($processorId)));
+        }
+
         switch ($handler['type']) {
         case 'service':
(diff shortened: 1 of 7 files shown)


DependencyInjection/MonologExtension.php
```diff
@@ -110,6 +110,16 @@ public function load(array $configs, ContainerBuilder $container)
         }
 
         $container->setParameter('monolog.additional_channels', isset($config['channels']) ? $config['channels'] : array());
+
+        if ($config['process_messages_using_psr_3_rules']) {
+            $processor = new Definition("Monolog\\Processor\\PsrLogMessageProcessor");
+            $processorId = uniqid('monolog.psr_log_message.processor.');
```
Seldaek: Why create one processor per handler that needs it? I think we could just have one fixed service id, as it is stateless  and configuration-less, a unique service can be injected in all handlers requiring it.

ged15: Fixed

ged15:
I thought about the design and I think it makes more sense to push the processor to each handler. That way, processing PSR-3 messages can be disabled per handler.

ged15:
Looks like this does not work for the `event` logger in Symfony. Will dig deeper to figure out why.

fabpot:
I've just tested and it looks good to me. @ged15 Can you rebase?

ged15:
@fabpot rebased and seems to work for me properly now too. Was a bug in an earlier version of the bundle I think.

fabpot:
Thanks, I'm going to merge this in the 2.x branch as this is both a new feature (BC) and a fix for the current behavior.

fabpot:
Thank you @ged15.

PR #175, opened 2016-09-29
Add Logmatic.io config

mRoca:
Add [Logmatic.io](https://logmatic.io/) configuration.

Caution : for the moment, the [LogmaticHandler](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/LogmaticHandler.php) is only present on Monolog master branch.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -281,6 +281,14 @@ use Monolog\Logger;
  *   - [level]: level name or int value, defaults to DEBUG
  *   - [bubble]: bool, defaults to true
  *
+ * - logmatic:
+ *   - token: Logmatic api token
+ *   - [hostname]: Host name supplied by Logmatic
+ *   - [app_name]: Application name supplied by Logmatic
+ *   - [use_ssl]: bool, defaults to true
+ *   - [level]: level name or int value, defaults to DEBUG
+ *   - [bubble]: bool, defaults to true
+ *
  * - server_log:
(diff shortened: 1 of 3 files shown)


lyrixx:
:+1:

fabpot:
... which means that it won't be easy to merge this one anytime soon as Monolog 2 beaks BC (which is fine).

stof:
this should indeed stay in stand-by until we figure out the way to support Monolog 2 (remembering we cannot drop support for Monolog 1 as we still need to support PHP 5.x until the end of the Symfony 3.4 LTS, and switching to 2.x only would make it hard for the ecosystem). The hard part is figuring out how to handle custom handlers shipped in the bridge actually (as this is where there is a BC break). This bundle deals with configuration, and adapting constructor arguments according to the version is quite easy to do.

alborq:
Up. Logmatic Handler has been release on Monolog.

stof:
@alborq that's not true. It exists only in the master branch, which is the dev version of Monolog 2.x

lyrixx:
Hello @mRoca 

Could you 
* rebase
* add a note in the CHANGELOG 
* throw an exception if the logmatic handler is configured but the class does not exist (1.X vs 2.X) 

Thanks

fabpot:
I think this one can be close as Logmatic does not exist anymore.

lyrixx:
Thanks again @mRoca and sorry for this delay

PR #176, opened 2016-10-03
Use "more entropy" option when using uniqid() function

javiereguiluz:
As you can see in [the source code of the PHP uniqid() function](https://github.com/php/php-src/blob/1c295d4a9ac78fcc2f77d6695987598bb7abcb83/ext/standard/uniqid.c#L68) when you don't pass `true` as the second argument, PHP sleeps the application for 1 microsecond (`usleep(1)`)

Symfony uses `true` almost everywhere, but this bundle still had some uniqid() calls without this option.


stof:
I was not aware of the usage of usleep in this case.

stof:
Thank you @javiereguiluz.

PR #177, opened 2016-10-18
Fix logger that logs to STDERR when no handlers are registered

fabpot:
Since Seldaek/monolog@ad37b7b, the `event` logger has no registered handlers in Symfony SE on the CLI. As Monolog 1.x automatically adds a STDERR handler when no handlers are registered, it means that logs (the event ones for instance) are output on STDERR by default with the Symfony CLI, which is exactly what we try to avoid.

So, this PR adds a new compiler pass to register a null handler when a logger has no registered handlers to avoid having the default STDERR handler from Monolog.

### DIFF
--- a/MonologBundle.php
+++ b/MonologBundle.php
@@ -20,6 +20,7 @@ use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Bundle\MonologBundle\DependencyInjection\Compiler\LoggerChannelPass;
 use Symfony\Bundle\MonologBundle\DependencyInjection\Compiler\DebugHandlerPass;
 use Symfony\Bundle\MonologBundle\DependencyInjection\Compiler\AddProcessorsPass;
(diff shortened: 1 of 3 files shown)


fabpot:
closes #177

stof:
Good catch, thanks @fabpot.

MonologBundle.php
```diff
@@ -34,6 +35,7 @@ public function build(ContainerBuilder $container)
 
         $container->addCompilerPass($channelPass = new LoggerChannelPass());
         $container->addCompilerPass(new DebugHandlerPass($channelPass));
+        $container->addCompilerPass(new RemoveEmptyLoggerPass($channelPass));
```
pierredup: Should this not be `FixEmptyLoggerPass`?

stof: Fixed in 0b42713

PR #180, opened 2016-10-19
Fix class name for FixEmptyLoggerPass in Bundle class

pierredup:


stof:
Already done when merging the PR.

PR #184, opened 2016-11-06
Deprecate DebugHandlerPass in favor of AddDebugLogProcessorPass in FrameworkBundle

nicolas-grekas:
Related to https://github.com/symfony/symfony/pull/20416

### DIFF
--- a/DependencyInjection/Compiler/DebugHandlerPass.php
+++ b/DependencyInjection/Compiler/DebugHandlerPass.php
@@ -11,6 +11,8 @@
 
 namespace Symfony\Bundle\MonologBundle\DependencyInjection\Compiler;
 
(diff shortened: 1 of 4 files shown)


fabpot:
Thank you @nicolas-grekas.

Tobion:
The `AddDebugLogProcessorPass` is only available in Symfony 3.2. Does it mean monolog-bundle 3.x is only compatible with Symfony >= 3.2?

xabbuh:
This (to be more precise, the changes done in #188) is indeed now breaking applications not using FrameworkBundle >= 3.2 (see symfony/symfony#21014 and https://github.com/symfony/monolog-bundle/pull/188#issuecomment-262410331). How do we deal with this now?

stof:
@nicolas-grekas can you explain how you planned to handle this ?

stof:
And this class cannot be deprecated, as it triggers deprecation for all Symfony LTS users

PR #185, opened 2016-11-07
Fix changelog version info

Tobion:
Those changes are in v3, not v2.


StevendeVries:
Any info on when v3 will be released?

Tobion:
It is already.

StevendeVries:
@Tobion Thanks, I was looking under releases instead of tags.

fabpot:
Thank you @Tobion.

PR #186, opened 2016-11-07
bumped version of PHP to the minimal one of Symfony 2.7

fabpot:
As suggested by @Tobion, I propose to update the minimal PHP version as 2.7 does not support old PHP 5.3 versions anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,6 @@ cache:
         - $HOME/.composer/cache/files
 
 php:
-  - 5.3
-  - 5.4
   - 5.5
   - 7.0
   - hhvm
@@ -16,7 +14,7 @@ php:
 matrix:
   fast_finish: true
   include:
-    - php: 5.3
+    - php: 5.5
       env: COMPOSER_FLAGS="--prefer-lowest" SYMFONY_DEPRECATIONS_HELPER=weak
     # Test against Symfony LTS versions
     - php: 5.6
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         }
     ],
     "require": {
-        "php": ">=5.3.2",
+        "php": ">=5.5.9",
         "symfony/monolog-bridge": "~2.7|~3.0",
         "symfony/dependency-injection": "~2.7|~3.0",
         "symfony/config": "~2.7|~3.0",


Tobion:
I actually meant PHP >= 5.5. I know it's compatible with Symfony 2.7 which is PHP 5.3, but that doesn't mean the library can increase the PHP independently. Otherwise you keep supporting an unsupported PHP version in a new release.

fabpot:
Bumped to 5.5. Ok with everybody?

nicolas-grekas:
:+1:

stof:
👍

stof:
Actually, there is a drawback: as the symfony.com download builds the deps for the min supported PHP version, it would mean that an outdated version of MonologBundle would be shipped in projects bootstrapped from the LTS. Not good for bug fixes.

stof:
Another solution is to change the symfony-standard so that the symfony.com archive builds deps for a newer PHP version than 5.3, acknowledging that PHP 5.3 is unmaintained since a long time). The symfony installer requires PHP 5.4+ anyway.

stof:
@fabpot what do you think about this ?

fabpot:
So, we would be building Symfony SE for PHP 5.5?

stof:
I would say yes. People really wanting to run their project on PHP 5.3 should run ``composer update`` to resolve dependencies compatible with their unmaintained PHP version.
Otherwise, we force people using maintained PHP versions to run a ``composer update`` to get uptodate dependencies.
I think the extra work should be for PHP 5.3 users, not for people using a maintained PHP version

stof:
The installer could display a warning when downloading an archive built with dependencies for a newer platform than the current one btw, as it still has access to ``config.platform.php`` in the composer.json (the installer is the one removing it)

Tobion:
Fully agree with @stof

stof:
@javiereguiluz any news about the installer improvements to unlock this ?

javiereguiluz:
There's no news. In any case, the Symfony Installer is completely deprecated now that we have Flex. I don't think well add any other feature or improvement to it. I'm sorry.

stof:
then it means we cannot merge this PR until the end of support of Symfony 2.8 without suffering from the drawback described in https://github.com/symfony/monolog-bundle/pull/186#issuecomment-260685538

Tobion:
If the installer is not maintained, I would also assume the pre-built symfony archieve is also not recommended anymore. So it does not matter if there is a slightly outdated version of the monolog bundle in it.

xabbuh:
I don't think adding such a warning would be a big issue.

ericpoe:
Since Symfony 2.8 requires at least PHP 5.4, can we at least drop PHP 5.3? Especially since 5.3 is no longer supported by Travis-CI on Trusty.

On Travis, PHP 5.3 requires Precise & HHVM requires Trusty. Unfortunately with travis-ci, we can't have it both ways. https://docs.travis-ci.com/user/languages/php

stof:
> Since Symfony 2.8 requires at least PHP 5.4

that's statement is wrong. It requires PHP 5.3.9+

> On Travis, PHP 5.3 requires Precise & HHVM requires Trusty

We can (as we don't need it in the same job). We do it for the Symfony repo

PR #191, opened 2016-11-29
Add SlackWebhookHandler and SlackbotHandler into configuration

jewome62:
Hello,

With the new version of Monolog, we have two new handler SlackWebhookHandler and SlackbotHandler 
https://github.com/Seldaek/monolog/pull/846

I purpose this PR for add it into monolog-bundle configuration.

I wait for your reviews if you need.

Thanks


Seldaek:
Thanks

Issue #165, opened 2016-05-01
Don't load DebugHandler by default


ostrolucky:
I consider big problem in Symfony ecosystem OOM errors for console commands. I dealt with it many times and lot of developers don't know the reason for this is DebugHandler, so they blame components which use monolog (e.g. doctrine). 

CLI commands are usually written for bulk tasks and since this bundle enables this handler by default, every action is saved to memory, even though the only Symfony Component which uses DebugHandler's content is Profiler, which we can't even use in CLI.

Of course, this can be prevented if we use prod environment or disable logging completely, but what if we still want logs written or outputted to terminal, but not saved in memory?

I think for saner defaults, this handler should be either:
- not autoloaded by default and let HttpKernel (or other Symfony component) to load it specifically, 
- not autoloaded by default in CLI
- autoloaded by default in HTTP only

Relevant: #37

Seldaek:
The defaults are handled in the https://github.com/symfony/symfony-standard repository actually. And IMO they are ok, as you shouldn't run prod jobs in dev environment. Not sure what else to say really.

gadelat:
No. Here are the defaults I'm talking about https://github.com/symfony/monolog-bundle/blob/master/MonologBundle.php#L35 What defaults are you talking about? If I run prod job in prod environment but I want to monitor SQL queries which doctrine is executing, I have to override those options to get rid of DebugHandler, otherwise I will end up with OOM. These are not good defaults considering DebugHandler is used normally only by symfony profiler.

xabbuh:
@gadelat But the `DebugHandlerPass` won't do anything if the `kernel.debug` container parameter is `false` (see https://github.com/symfony/monolog-bundle/blob/master/DependencyInjection/Compiler/DebugHandlerPass.php#L41-L43) which is the case when running the application in the prod environment.

gadelat:
1. I'm saying this now for 3rd time. Sometimes you want to see debugging info even on production/with production data, or the task you are writing just naturally produces lot of logs and you want to see them. In that case, you can't set `kernel.debug` to false. Every bundle uses this for defaults. Almost nothing would get logged in that case, unless you override this in config for every component which uses monolog. So, if I want to debug my large batch task, I would need to either override this in config for every component I need logs from, or write my task with this in mind so it manipulates monolog handlers by itself. Both of them are workarounds. Personally I would rather use the third, dirty approach and just temporarily comment out the DebugHandler from MonologBundle - much less work.
[...]

PR #195, opened 2016-12-27
revert disabling DebugHandler in CLI environments

xabbuh:
As the cache is shared between the different server APIs, the debug handler needs to take care of abstaining itself.

After symfony/symfony#21061 being merged, the changes made in fbbcefd7821edfaa31a48712bba0960979243c1c should no longer be necessary.

### DIFF
--- a/DependencyInjection/Compiler/DebugHandlerPass.php
+++ b/DependencyInjection/Compiler/DebugHandlerPass.php
@@ -46,11 +46,6 @@ class DebugHandlerPass implements CompilerPassInterface
             return;
         }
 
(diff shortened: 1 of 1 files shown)


xabbuh:
Please not that both PRs do not solve https://github.com/symfony/monolog-bundle/issues/165#issuecomment-262452076. For that we would probably have to introduce a new config option in MonologBundle that would allow to customise the `$skipOnCli` property introduced in symfony/symfony#21061.

stof:
I'm merging this because the PR broke cache warmup

stof:
Thank you @xabbuh.

PR #196, opened 2016-12-28
Added support for ServerLog command

lyrixx:
Support for https://github.com/symfony/symfony/pull/21080

### DIFF
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -677,6 +677,17 @@ class MonologExtension extends Extension
                 $handler['app_name'],
             ));
             break;
+        case 'server_log':
+            if (!class_exists('Symfony\Bridge\Monolog\Handler\ServerLogHandler')) {
+                throw new \RuntimeException('The ServerLogHandler is not available. Please update "symfony/monolog-bridge" to 3.3.');
+            }
+
+            $definition->setArguments(array(
+                $handler['host'],
+                $handler['level'],
+                $handler['bubble'],
+            ));
+            break;
 
         // Handlers using the constructor of AbstractHandler without adding their own arguments
         case 'browser_console':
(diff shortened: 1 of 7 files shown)


DependencyInjection/MonologExtension.php
```diff
@@ -652,6 +652,13 @@ private function buildHandler(ContainerBuilder $container, $name, array $handler
                 $handler['app_name'],
             ));
             break;
+        case 'server_log':
+            $definition->setArguments(array(
+                $handler['host'],
```
xabbuh: In the `ServerLogHandler` class, the argument is named `$connectionString`. I would unify that.

xabbuh: And there should better be a nice exception being thrown if the new handler is not present.

xabbuh:
The new handler option (whatever its final name) must be added to the XSD file.

DependencyInjection/MonologExtension.php
lyrixx: I renamed it `host` in symfony.

lyrixx: > And there should better be a nice exception being thrown if the new handler is not present.

It can occur if one use the very lastest version of the bundle, but with an older version symfony?

xabbuh: > It can occur if one use the very lastest version of the bundle, but with an older version symfony?

Exactly as the release cycle of MonologBundle is independent of Symfony's release cycle.

lyrixx:
> The new handler option (whatever its final name) must be added to the XSD file.

It looks like the .xsd is really behind the Configuration. Are you sure I need to add something there?

xabbuh:
> It looks like the .xsd is really behind the Configuration. Are you sure I need to add something there?

IMO yes, if there is anything else missing, we should sync the XSD file in a separate PR.

lyrixx:
@xabbuh I added a tests and it looks like I don't need to updated the xsd.

lyrixx:
\o/ Finally green on all versions.

Tests/DependencyInjection/Fixtures/yml/server_log.yml
```diff
@@ -0,0 +1,5 @@
+monolog:
+    handlers:
+        server_log:
+            type:       server_log
+            host:       0:9911
```
Seldaek: Please add this as an example on top of the Configuration class (in the class docblock we have examples of all handlers)

lyrixx: done ;)

PR #198, opened 2017-01-02
Revert "Drop DebugHandlerPass"

### LINKED ISSUE #190: After updating to version 3.0.1 no logs in profiler
After updating this package from 2.11.1 to 3.0.1 logs stops showing in web profiler. 

I use Symfony 3.1.6

screen

stof:
This reverts commit b46da89f264efa7ed42ed3db6b9b69a2b2ef2163.
The class is necessary to support Symfony <3.2. Closes #190.

The 3.x version of the bundle contains only 2 changes compared to 2.11:

- breaking support for Symfony <3.2 (while still announcing compatibility), which is what got reverted here
- adding support for a new Monolog feature (which could totally fit in 2.x btw).

As we still need to keep support for the Symfony LTS, and the 3.x series is currently broken for us, I'm suggesting here to revert the removal (effectively making it a follow-up of 2.x, which would allow to avoid maintaining 2 branches).
The other alternative would be to kill the existing 3.x releases entirely to forbid them to be installed in a Symfony 2.8 project allowing ``symfony/monolog-bundle: ^2.8 || ^3``, but composer does not support yanking existing tags, and git is bad at deleting them (deleting them would require deleting them in all existing clones to be sure that they would never be pushed again to the repo, and then deleting the Packagist versions).

### DIFF
--- a/DependencyInjection/Compiler/DebugHandlerPass.php
+++ b/DependencyInjection/Compiler/DebugHandlerPass.php
@@ -1,0 +1,63 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MonologBundle\DependencyInjection\Compiler;
+
+@trigger_error('The '.__NAMESPACE__.'\DebugHandlerPass class is deprecated since version 2.12 and will be removed in 3.0. Use AddDebugLogProcessorPass in FrameworkBundle instead.', E_USER_DEPRECATED);
+
+use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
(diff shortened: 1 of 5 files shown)


nicolas-grekas:
:+1:
`DebugHandlerPass` should have been removed before tagging 3.0. Since we missed this step, we have to postpone the removal for 4.0.
@stof can you update the deprecation message thought?

stof:
@nicolas-grekas you did remove it in 3.0, which is exactly what broke support for Symfony <3.2 (without changing the composer constraint), as the suggested replacement is a Symfony 3.2 picture.

@nicolas-grekas regarding the deprecation warning, I will update it separately (after merging the revert of the removal), as it should be done in 2.x instead

nicolas-grekas:
We did not, that's what prevents a proper fix here, see:
https://github.com/symfony/monolog-bundle/blob/3.0.0/DependencyInjection/Compiler/DebugHandlerPass.php

stof:
oh, that's even worse, 3.0 is BC with 2.x (with more features I forgot to list above), and 3.0.1 broke BC and Symfony support.
So the revert is indeed the only meaningful way. but in any case, the removal without bumping the Symfony constraint was wrong.

hvt:
Would be great @stof if you could tag a new release including this fix.

stof:
@hvt I'm on it

stof:
@hvt releases are done

PR #203, opened 2017-02-03
Add autowiring alias for LoggerInterface

chalasr:
Fixes https://github.com/symfony/symfony/pull/21517#issuecomment-277216071

### DIFF
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -52,12 +52,11 @@ class MonologExtension extends Extension
         if (isset($config['handlers'])) {
             $loader = new XmlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
             $loader->load('monolog.xml');
-            $container->setAlias('logger', 'monolog.logger');
 
             $container->setParameter('monolog.use_microseconds', $config['use_microseconds']);
 
-            // always autowire the main logger, require Symfony >= 2.8
-            if (method_exists('Symfony\Component\DependencyInjection\Definition', 'addAutowiringType')) {
+            // always autowire the main logger, require Symfony >= 2.8, < 3.3
+            if (!method_exists('Symfony\Component\DependencyInjection\ContainerBuilder', 'fileExists') && method_exists('Symfony\Component\DependencyInjection\Definition', 'addAutowiringType')) {
                 $container->getDefinition('monolog.logger')->addAutowiringType('Psr\Log\LoggerInterface');
             }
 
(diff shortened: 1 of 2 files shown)


DependencyInjection/MonologExtension.php
```diff
@@ -59,6 +60,8 @@ public function load(array $configs, ContainerBuilder $container)
             // always autowire the main logger, require Symfony >= 2.8
             if (method_exists('Symfony\Component\DependencyInjection\Definition', 'addAutowiringType')) {
                 $container->getDefinition('monolog.logger')->addAutowiringType('Psr\Log\LoggerInterface');
+            } else {
```
stof: that's wrong. Symfony 3.3 does have the method, so it will never enter the ``else`` part, while it is the version where you want the alias.

currently, you only define the alias for Symfony < 2.8, where it is useless

chalasr:
@stof fixed, thanks

DependencyInjection/MonologExtension.php
```diff
@@ -57,6 +58,7 @@ public function load(array $configs, ContainerBuilder $container)
             $container->setParameter('monolog.use_microseconds', $config['use_microseconds']);
 
             // always autowire the main logger, require Symfony >= 2.8
+            $container->setAlias('Psr\Log\LoggerInterface', new Alias('monolog.logger', false));
```
stof: I would put this directly in the XML file.

Btw, I would put the autowiring point to ``logger`` itself (making it easier for people replacing the main logger service)

chalasr: Done

nicolas-grekas:
the next nice thing would be to not register the autowiring type when not required (prevents a deprecation)

chalasr:
@nicolas-grekas done

chalasr:
hmm checking the kernel version doesn't work since it doesn't guarantee the DI version. Any clue?

chalasr:
@nicolas-grekas done

DependencyInjection/MonologExtension.php
```diff
@@ -57,7 +57,7 @@ public function load(array $configs, ContainerBuilder $container)
             $container->setParameter('monolog.use_microseconds', $config['use_microseconds']);
 
             // always autowire the main logger, require Symfony >= 2.8
```
xabbuh: the comment should be updated too

chalasr: done

nicolas-grekas:
:+1:

Seldaek:
Merging then assuming it's fine.. I haven't looked in details.

PR #207, opened 2017-03-08
Remove faulty service definitions

### LINKED ISSUE #192: monolog.activation_strategy.not_found gets created without parameters inside container
I have just upgraded my symfony/symfony installation to 3.2 and symfony/monolog-bundle to 3.0.1 and my smoke tests fail when they try to access the following services:

[monolog.activation_strategy.not_found](https://github.com/symfony/monolog-bundle/blob/master/Resources/config/monolog.xml#L21) and [monolog.handler.fingers_crossed.error_level_activation_strategy](https://github.com/symfony/monolog-bundle/blob/master/Resources/config/monolog.xml#L22)

The error I'm getting is the following:

```
Catchable Fatal Error: 
Argument 1 passed to Symfony\Bridge\Monolog\Handler\FingersCrossed\NotFoundActivationStrategy::__construct() 
must be an instance of Symfony\Component\HttpFoundation\RequestStack, none given, called
 in /project/var/cache/dev/appDevDebugProjectContainer.php on line 2769 and defined
```
and when I go and see my `appTestDebugProjectContainer.php` file I see these services have no arguments in their construction

```
    /**
     * Gets the 'monolog.activation_strategy.not_found' service.
     *
     * This service is shared.
     * This method always returns the same instance of the service.
     *
     * @return \Symfony\Bridge\Monolog\Handler\FingersCrossed\NotFoundActivationStrategy A Symfony\Bridge\Monolog\Handler\FingersCrossed\NotFoundActivationStrategy instance
     */
    protected function getMonolog_ActivationStrategy_NotFoundService()
    {
        return $this->services['monolog.activation_strategy.not_found'] = new \Symfony\Bridge\Monolog\Handler\FingersCrossed\NotFoundActivationStrategy();
    }
```

```
    /**
     * Gets the 'monolog.handler.fingers_crossed.error_level_activation_strategy' service.
     *
     * This service is shared.
     * This method always returns the same instance of the service.
     *
     * @return \Monolog\Handler\FingersCrossed\ErrorLevelActivationStrategy A Monolog\Handler\FingersCrossed\ErrorLevelActivationStrategy instance
     */
    protected function getMonolog_Handler_FingersCrossed_ErrorLevelActivationStrategyService()
    {
        return $this->services['monolog.handler.fingers_crossed.error_level_activation_strategy'] = new \Monolog\Handler\FingersCrossed\ErrorLevelActivationStrategy();
    }
```

michaelcullum:
These definitions have been broken for a while now with the only people
noticing being those running smoke tests on their DICs so I'd suggest
they can actually be removed instead of fixed.

Closes #192

### DIFF
--- a/Resources/config/monolog.xml
+++ b/Resources/config/monolog.xml
@@ -18,9 +18,6 @@
             <argument /><!-- Channel -->
         </service>
 
-        <service id="monolog.activation_strategy.not_found" class="Symfony\Bridge\Monolog\Handler\FingersCrossed\NotFoundActivationStrategy"/>
-        <service id="monolog.handler.fingers_crossed.error_level_activation_strategy" class="Monolog\Handler\FingersCrossed\ErrorLevelActivationStrategy"/>
-
         <!-- Formatters -->
         <service id="monolog.formatter.chrome_php" class="Monolog\Formatter\ChromePHPFormatter" public="false" />
         <service id="monolog.formatter.gelf_message" class="Monolog\Formatter\GelfMessageFormatter" public="false" />


stof:
I think these were meant to be used as parent definitions instead. So to avoid BC breaks there, I would mark them as abstract and deprecated.

enumag:
Looks like this blocks Symfony 3.3 compatibility.

```                                                    
  [Symfony\Component\DependencyInjection\Exception\RuntimeException]           
  Cannot autowire service "monolog.activation_strategy.not_found": argument "  
  $excludedUrls" of method "Symfony\Bridge\Monolog\Handler\FingersCrossed\Not  
  FoundActivationStrategy::__construct()" must have a type-hint or be given a  
   value explicitly.
```

stof:
This looks weird to me. These services are not opting into autowiring. Looks like the autowiring pass might be buggy when deciding whether it should process a service or no

enumag:
@stof I'm using https://github.com/Symplify/DefaultAutowire, maybe that causes the problem.

stof:
yeah, if you have a bundle turning on autowiring for other bundles which have not used it, it will indeed break things. This is precisely why Symfony does not force all services to be autowired. This bundle should die.

stof:
Btw, on Symfony 3.3, use file-level defaults to avoid having to configure each services as autowired (which would remove the need for this bundle entirely)

enumag:
@stof It works just fine with all bundler except this one. The thing is these two services are not properly configured and the compiled code for them would result in error anyway. Symfony/DependencyInjection really should be more strict about these things and throw an exception in complile time like Nette/DI would.

stof:
@enumag I agree that these service definitions are buggy. But the bundle forcing all services to be autowired is still a bad thing (other things might break, as it might change the service configuration compared to what the bundle provides)

enumag:
@stof I don't disagree. However the worst problem here is that Symfony allows these incompletes services to exist in the first place.

michaelcullum:
@stof I'd disagree on the 'fix and deprecate approach' as ultimately they don't work in their current state and therefore cannot have been being used by anyone. You can't deprecate something that never worked (and was therefore in place?) in the first place, and fixing something that never worked and deprecating it simultaneously seems superfluous.

stof:
@michaelcullum they can still be used as parent services (even though it is quite useless as they define only the class, not additional config, so defining the class on your service is not harder than using inheritance). This is why I want to make them abstract (to get a better error message when trying to use them the wrong way) and deprecated (because of no need) without breaking the working case.

xabbuh:
Can be closed now that #256 is merged?

Issue #208, opened 2017-03-10
Slackbot handler does not work


melnikoved:
Slackbot handler does not work, because there is grammar mistake and curl can not handle `#` sign from channel name.

PR #209, opened 2017-03-10
bug  #208 Slackbot handler does not work

melnikoved:
Slackbot handler does not work, because there is grammar mistake and curl can not handle # sign from channel name.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -219,7 +219,7 @@ use Monolog\Logger;
  *   - [bubble]: bool, defaults to true
  *
  * - slackbot:
- *   - team: slack team slug
+ *   - slack_team: slack team slug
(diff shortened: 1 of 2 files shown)


DependencyInjection/Configuration.php
```diff
@@ -219,7 +219,7 @@
  *   - [bubble]: bool, defaults to true
  *
  * - slackbot:
- *   - team: slack team slug
+ *   - slack_team: slack team slug
```
lyrixx: I would made the change in the opposite way to preserve the BC: Just update the Monolog Extension.

melnikoved:
OK. I made changes. 

Best regards, 
Eduard Melnikov

melnikoved:
And there was the mistake:  stack_team instead of team

Best regards, 
Eduard Melnikov

PR #210, opened 2017-03-21
Allow to override VERBOSITY_QUIET in configuration

mweibel:
VERBOSITY_QUIET is set to ERROR by default and you can't override that, even though the OutputInterface allows to use this verbosity. In the default configuration, quiet doesn't mean really quiet as errors are still logged.


Tobion:
👍

Status: Reviewed

Since https://github.com/symfony/symfony/pull/18781 the quite mode is actually broken because it now also outputs debug messages as this is the default when no configuration is set. So this PR is actually important to fix this new behavior again.

PR #211, opened 2017-03-22
Add autowiring aliases

nicolas-grekas:


chalasr:
https://github.com/symfony/monolog-bundle/pull/203 ? :)

DependencyInjection/MonologExtension.php
```diff
@@ -52,7 +52,6 @@ public function load(array $configs, ContainerBuilder $container)
         if (isset($config['handlers'])) {
             $loader = new XmlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
             $loader->load('monolog.xml');
-            $container->setAlias('logger', 'monolog.logger');
```
nicolas-grekas: @chalasr can you add this patch to your PR?

chalasr: @nicolas-grekas done

PR #215, opened 2017-05-10
Combine multiple isset() calls when possible

javiereguiluz:
`isset($a) && isset($b)` is the same as `isset($a, $b)`, so we can save 1 function call. We did the same for Symfony: https://github.com/symfony/symfony/pull/22684 and https://github.com/symfony/symfony/pull/22685


stof:
See my comment on https://github.com/symfony/symfony/pull/22684

PR #222, opened 2017-08-21
Add formatter_options to console handler

grongor:
This patch is only applicable if this PR https://github.com/symfony/symfony/pull/23929 gets merged.

I'm not exactly sure about the changes in `Configuration.php`, so please take a closer look and let me eventually know if I should change something. Thanks, cheers!


lyrixx:
Hello, I think  this issue is solved thanks to #30345 and symfony/monolog-bundle#297

@grongor Could you confirm it's OK?

grongor:
Fixed by https://github.com/symfony/symfony/pull/23929

Issue #223, opened 2017-08-22
Failure on count(string) in 7.2 for fingers crossed config


iltar:
In php 7.2, you can no longer count strings, which triggers a notice. This notice being (rightfully) converted into an exception by symfony in dev/test. https://3v4l.org/qILT9

For some reason, the check here checks if the count on the string (the handler, which is `'buffered'`) equals 1. This is true so it works. However, this will become deprecated behavior. I'm just not sure what the correct check would be to make a PR to fix this.

```
Fatal error: Uncaught Symfony\Component\Debug\Exception\ContextErrorException: Warning: count(): Parameter must be an array or an object that implements Countable in /home/ivanderberg/projects/mijn/www/vendor/symfony/symfony/src/Symfony/Component/Config/Definition/BaseNode.php on line 313

Symfony\Component\Config\Definition\Exception\InvalidConfigurationException: Invalid configuration for path "monolog.handlers.mail": Warning: count(): Parameter must be an array or an object that implements Countable in /home/ivanderberg/projects/mijn/www/vendor/symfony/symfony/src/Symfony/Component/Config/Definition/BaseNode.php on line 313

Call Stack:
    0.7772   16931896   1. Symfony\Component\Debug\ErrorHandler->handleException() /home/ivanderberg/projects/mijn/www/vendor/symfony/symfony/src/Symfony/Component/Debug/ErrorHandler.php:0

```

```yml
monolog:
    handlers:
        streamed:
            type         : rotating_file
            max_files    : 5
            path         : "%kernel.logs_dir%/%kernel.environment%.log"
            level        : debug
            channels     : ["!event"]
        mail:
            type         : fingers_crossed
            action_level : error
            handler      : buffered
        buffered:
            type         : buffer
            handler      : swift
        swift:
            type         : swift_mailer
            from_email   : "%mailing.error_mail%"
            to_email     : "%mailing.error_mail%"
            subject      : "%mailing.subject.error%"
            level        : info
            channels     : ["!event"]
```

The following snippet ([source](https://github.com/symfony/monolog-bundle/blob/master/DependencyInjection/Configuration.php#L653-L656)):
```php
->validate()
    ->ifTrue(function ($v) { return (
            'fingers_crossed' === $v['type'] || 
            'buffer' === $v['type'] || 
            'filter' === $v['type']
        ) && 1 !== count($v['handler']); })
[...]

PR #224, opened 2017-09-09
bug #223 Failure on count(string) in 7.2 for fingers crossed config

VolCh:
Check that `hander` is not empty instead of use the old PHP behavior `count('string') === 1`

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -651,7 +651,7 @@ class Configuration implements ConfigurationInterface
                             ->thenInvalid('Service handlers can not have a formatter configured in the bundle, you must reconfigure the service itself instead')
                         ->end()
                         ->validate()
-                            ->ifTrue(function ($v) { return ('fingers_crossed' === $v['type'] || 'buffer' === $v['type'] || 'filter' === $v['type']) && 1 !== count($v['handler']); })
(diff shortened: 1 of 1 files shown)


DependencyInjection/Configuration.php
```diff
@@ -651,7 +651,7 @@ public function getConfigTreeBuilder()
                             ->thenInvalid('Service handlers can not have a formatter configured in the bundle, you must reconfigure the service itself instead')
                         ->end()
                         ->validate()
-                            ->ifTrue(function ($v) { return ('fingers_crossed' === $v['type'] || 'buffer' === $v['type'] || 'filter' === $v['type']) && 1 !== count($v['handler']); })
+                            ->ifTrue(function ($v) { return ('fingers_crossed' === $v['type'] || 'buffer' === $v['type'] || 'filter' === $v['type']) && (empty($v['handler']) || !is_string($v['handler'])); })
```
nicolas-grekas: shouldn't it be just `empty($v['handler'])`? that would match the other checks around, and match the intent, isn't it?

VolCh: You are right. Changed.

hboomsma:
Any news on this issue?

tarlepp:
Any updates with this issue?

pascaldevink:
Any updates on this? It is keeping us from further testing PHP 7.2 with our application.
If the failing build is the only thing keeping it from being merged: it looks like none of the failures are related to this change, so maybe a rebase with master might do the trick?

VolCh:
May still need some action on my part?

PR #233, opened 2017-10-03
Remove extra spaces

ahmedprotm:
remove extra space from MonologExtension.php


fabpot:
The merge commit should be removed by rebasing on current master.

PR #235, opened 2017-10-20
Fix invalid use of count() in PHP 7.2

merk:
ref https://github.com/symfony/monolog-bundle/issues/223

PHP 7.2+ issues a warning when trying to count a non Iterable. The case in this code is not clear if it should just check for the presence of a handler or not. I suspect the original intention was to make sure the specified handler is defined in the configuration, but this should happen in the Extension?


VolCh:
I made PR with the same change more one month ago https://github.com/symfony/monolog-bundle/pull/224

fabpot:
Thank you @merk.

PR #238, opened 2017-11-06
use symfony/phpunit-bridge for tests

fabpot:


.travis.yml
```diff
@@ -5,6 +5,7 @@ language: php
 cache:
     directories:
         - $HOME/.composer/cache/files
+        - .phpunit
```
xabbuh: The bridge will not use this directory by default, will it? IIRC we need to use the `PHPUNIT_DIR` env var to point to this directory.

composer.json
```diff
@@ -26,6 +26,7 @@
     "require-dev": {
         "symfony/yaml": "~2.3|~3.0|~4.0",
         "symfony/console": "~2.3|~3.0|~4.0",
+        "symfony/phpunit-bridge": "^3.3",
```
xabbuh: `^3.3|^4.0`?

composer.json
```diff
@@ -26,6 +26,7 @@
...
         "phpunit/phpunit": "^4.8"
```
xabbuh: We could drop this dependency then, right?

fabpot: indeed

composer.json
fabpot: done

.travis.yml
fabpot: done

.travis.yml
```diff
@@ -26,6 +26,9 @@ matrix:
     # Test against dev versions
     - php: 5.6
```
xabbuh: Should we change this to PHP 7? Or add an explicit job for Symfony 4?

.travis.yml
```diff
@@ -26,6 +26,9 @@ matrix:
...
       env: DEPENDENCIES=dev
+    - php: 7.0
+    - php: 7.1
+    - php: hhvm
```
xabbuh: Is this still required? We already dropped it for Symfony, didn't we?

.travis.yml
stof: I agree about changing this to PHP 7.1 (dev deps should be tested against a recent PHP version).

And we should add a LTS job for 3.4 (similar to what is done for 2.8, which should not be using dev deps anymore btw)

.travis.yml
```diff
@@ -37,4 +40,4 @@ before_install:
 install:
     - composer update $COMPOSER_FLAGS
```
stof: can you add `./vendor/bin/simple-phpunit install` in this section too, so that the installation output from simple-phpunit gets collapsed by default on Travis (currently, the installation is done by the actual run)

PR #240, opened 2017-11-21
Remove Extra spaces

adouiri:
Remove Extra spaces


derrabus:
Same as #233.

Tobion:
Good catch, thanks @adouiri.

PR #242, opened 2017-12-07
Add stream handler lock option

mRoca:
This PR adds the $useLocking StreamHandler argument : https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/StreamHandler.php#L44


zerkms:
@Seldaek @stof @fabpot team, can we please merge it? It really is essential

lyrixx:
It took time, but here we go, this is in now. Thank you very much @mRoca.

Issue #220, opened 2017-06-26
It is not possible to use ENV variable for log level


asp24:
Due to https://github.com/symfony/monolog-bundle/blob/master/DependencyInjection/MonologExtension.php#L37

it is not possible to specify log level using '%env(LOG_LEVEL%)' string as value

```yaml
monolog:
    handlers:
        main:
            type:         'fingers_crossed'
            action_level: '%env(LOG_LEVEL)%'
            handler:      'nested'
            buffer_size:  10
        nested:
            type:  'stream'
            path:  '%kernel.logs_dir%/%kernel.environment%.log'
            level: 'debug'
        console:
            type:  'console'
```

```
PHP Fatal error:  Uncaught Symfony\Component\Debug\Exception\ContextErrorException: Warning: constant(): Couldn't find constant Monolog\Logger::%env(LOG_LEVEL)% in ./vendor/symfony/monolog-bundle/DependencyInjection/MonologExtension.php:42
```

ababushkin:
Any update on this one?

alterphp:
Going to be fixed at Symfony level (many other bundles configs have the same problem) with https://github.com/symfony/symfony/pull/23888

stof:
this particular case will *not* be fixed by the Symfony change (turning the config into the constant value simply cannot be done without having the *value* of the env var)

alterphp:
And what about an intermediate parameter as following : 
```

**Comment by alterphp** [comment]
@stof  And what about an intermediate parameter ?

```
parameters:
    monolog.main.level: '%env(MONOLOG_LOG_LEVEL)%'
monolog:
    handlers:
        main:
            type:         'fingers_crossed'
            action_level: '%monolog.main.level%'
```

Will the symfony PR fix that case ?

stof:
Nope, because the parameters gets resolved already, so this is exactly the same case than using the env placeholder directly. In any case, the conversion of the constant name to the constant value **cannot** work when having a placeholder instead of the actual config.

The Symfony change will make things work in places performing a type validation (so places enforcing an integer will accept the placeholder if it is a `%env(int:...)%` placeholder casting to int). But it cannot make complex transformations work.

alterphp:
Ok @stof , thanks for details.

How do you handle multiple stages log level configuration at Sensio ? Is the production value fixed in the config ?

ClementGautier:
Hello there ! 

So what the best workaround in your opinion guys ?
[...]

PR #245, opened 2018-01-11
Allow environment variables in log level configuration.

umulmrum:
This PR allows to use environment variables for log levels, e.g. "%env(APP_LOGLEVEL)%", as requested in #220.
The log level can be either given as an int (the constant values in Monolog\Logger) or as a string (the constant names in Monolog\Logger). For the latter case, an EnvVarProcessor with prefix "loglevel" was added, which converts constant names to int values.

Examples: 

1. When using a variable in the form "%env(APP_LOGLEVEL)%", an int needs to be passed from the environment.
2. When using a variable in the form "%env(loglevel:APP_LOGLEVEL)%", a string needs to be passed from the environment.

The test for the console logger was not changed, as I couldn't figure out how to configure verbosity_levels in XML. What makes this also difficult is that environment variable configs are replaced during container build, using some random suffixes (https://github.com/symfony/symfony/pull/19681), so that it might not be possible to test the Configuration class reliably. Of course, input on this question is appreciated.

edit: D'oh, I forgot that this change is incompatible with older Symfony versions. I'll try to solve this. edit: \o/


lyrixx:
Replaced by #357

PR #246, opened 2018-01-15
Add ignore_error option to gelf handler

paxal:
`ignore_error` option to gelf handler wraps the transport inside the provided [`Gelf\Transport\IgnoreErrorTransportWrapper`](https://github.com/bzikarsky/gelf-php/blob/152c155df9795d62a096fd134e43671c13cc0e37/src/Gelf/Transport/IgnoreErrorTransportWrapper.php) added in graylog2/gelf-php:1.5 release.

To respect default behavior, it is set to `false` by default.

Thus, one can configure the gelf handler using basic configuration without the need to create a dedicated service.

The same option seems to exist for [elasticsearch](https://github.com/symfony/monolog-bundle/blob/2b41b8b6d2c6edb1a5494f02f8e4129be2a44784/DependencyInjection/Configuration.php#L475) [handler](https://github.com/symfony/monolog-bundle/blob/2b41b8b6d2c6edb1a5494f02f8e4129be2a44784/DependencyInjection/MonologExtension.php#L292) but I'm not sure it does the same thing (:O

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -46,7 +46,7 @@ use Monolog\Logger;
  *   - [bubble]: bool, defaults to true
  *
  * - gelf:
- *   - publisher: {id: ...} or {hostname: ..., port: ..., chunk_size: ...}
+ *   - publisher: {id: ...} or {hostname: ..., port: ..., chunk_size: ..., ignore_error: ...}
  *   - [level]: level name or int value, defaults to DEBUG
  *   - [bubble]: bool, defaults to true
  *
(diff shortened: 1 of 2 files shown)


DependencyInjection/Configuration.php
```diff
@@ -413,6 +413,7 @@ public function getConfigTreeBuilder()
                                     ->scalarNode('hostname')->end()
                                     ->scalarNode('port')->defaultValue(12201)->end()
                                     ->scalarNode('chunk_size')->defaultValue(1420)->end()
+                                    ->booleanNode('ignore_error')->defaultValue(false)->end()
```
Tobion: logging errors should never break the app , esp. in production. @Seldaek does monolog (bundle) not catch processing errors automatically? If not, shouldn't handler error be catched always? This is similar to caching where the PSR says, caching errors should never throw exceptions.

Tobion: And if this get's only handled on a handler-by-handler basis, this ignore_error option should IMO default to `%kernel.debug%`.

Seldaek: The best "monolog" way to handle this is to wrap handlers in a WhatFailureGroupHandler which suppresses any exception.. That can apply to any handler instead of being gelf-specific

PR #247, opened 2018-01-24
Symfony 4 compatibility

romainneutron:
Hello,

This PR finishes the work started in #241. After asking @gnugat, he allowed me to address @Tobion comment (https://github.com/symfony/monolog-bundle/pull/241#pullrequestreview-84019901)

Comments are welcome

### DIFF
--- a/DependencyInjection/Compiler/LoggerChannelPass.php
+++ b/DependencyInjection/Compiler/LoggerChannelPass.php
@@ -113,6 +113,7 @@ class LoggerChannelPass implements CompilerPassInterface
                 $logger = new DefinitionDecorator('monolog.logger_prototype');
             }
 
+            $logger->setPublic(true);
             $logger->replaceArgument(0, $channel);
             $container->setDefinition($loggerId, $logger);
             $this->channels[] = $channel;
--- a/Resources/config/monolog.xml
+++ b/Resources/config/monolog.xml
@@ -12,16 +12,16 @@
             </call>
         </service>
 
-        <service id="logger" alias="monolog.logger" />
+        <service id="logger" alias="monolog.logger" public="true" />
 
         <service id="Psr\Log\LoggerInterface" alias="logger" public="false" />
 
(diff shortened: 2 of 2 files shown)


DependencyInjection/Compiler/LoggerChannelPass.php
```diff
@@ -113,6 +113,7 @@ protected function createLogger($channel, $loggerId, ContainerBuilder $container
                 $logger = new DefinitionDecorator('monolog.logger_prototype');
             }
 
+            $logger->setPublic(true);
```
stof: IMO, we should make them public only if they are channels requested to exist by the configuration of the bundle, not just by other services requesting a specific channel in a tag (as such case does not get the logger through `Container::get`

weaverryan: In general, making things public I think is a dying solution. If you use AbstractController, you can’t even get public services. I’m not against this change, but more broadly, we need to use DI. And if that’s too hard in some cases, we need to make it easier.

In this case, the best solution might be to add a binding for all you services (so, under _defaults) for your custom channel - e.g. if you create “api”, then you setup a $apiLogger as a bind to this service. Then you can get it anywhere by using that as its argument name.

@nicolas-grekas sorry to bother you, but this is what I was talking to you about a few days ago: where private services are a pain, but by using binding (if we like that solution), you can solve it easily.

Tobion:
So how about we let loggers go private in sf 4 and don't do anything? No need to make this more complicated if we don't really want them public anyway.

weaverryan:
I’d be fine with that. W can document using bind under _defaults to create project conventions for easy injection:

bind:
    $myLogger: ‘@monolog.logger.my_logger’

Tobion:
Let's close then. I'm surprised there haven't been more reports about private loggers. So maybe people don't need this anymore.

lyrixx:
Yes we still need it. But instead we are injecting it via the constructor. So we slowly migration to controller as a service because the framework for us to :( 

Note: I'm not saying controller as a service are bad, it's just not cool to be forced to migrate lot of code when updating an application.

PR #248, opened 2018-02-07
Remove randomness from container compilation

derrabus:
Following @Tobion's comment https://github.com/symfony/symfony/issues/25958#issuecomment-363462107, I've made an attempt to eliminate all randomly generated service IDs from `MonologExtension`.

`MonologExtension` currently uses `uniqid` to generate a service ID to be used in a single reference afterwards. I'm now directly inlining the affected definitions, so there is not need for randomly generated service IDs anymore.

### DIFF
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -184,22 +184,18 @@ class MonologExtension extends Extension
 
         case 'gelf':
             if (isset($handler['publisher']['id'])) {
-                $publisherId = $handler['publisher']['id'];
+                $publisher = new Reference($handler['publisher']['id']);
             } elseif (class_exists('Gelf\Transport\UdpTransport')) {
                 $transport = new Definition("Gelf\Transport\UdpTransport", array(
                     $handler['publisher']['hostname'],
                     $handler['publisher']['port'],
                     $handler['publisher']['chunk_size'],
                 ));
-                $transportId = uniqid('monolog.gelf.transport.', true);
                 $transport->setPublic(false);
-                $container->setDefinition($transportId, $transport);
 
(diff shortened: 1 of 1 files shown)


DependencyInjection/MonologExtension.php
```diff
@@ -184,46 +184,40 @@ private function buildHandler(ContainerBuilder $container, $name, array $handler
...
             if (isset($handler['publisher']['id'])) {
-                $publisherId = $handler['publisher']['id'];
+                $publisher = new Reference($handler['publisher']['id']);
             } elseif (class_exists('Gelf\Transport\UdpTransport')) {
                 $transport = new Definition("Gelf\Transport\UdpTransport", array(
                     $handler['publisher']['hostname'],
                     $handler['publisher']['port'],
                     $handler['publisher']['chunk_size'],
                 ));
-                $transportId = uniqid('monolog.gelf.transport.', true);
```
Tobion: some other handlers build the transportId as a sha hash from the config. I think this could be done here as well. This way, it would reuse the same UdpTransport for several gelf handlers when the hostname and port are the same. So it would only create one stream socket, see https://github.com/bzikarsky/gelf-php/blob/master/src/Gelf/Transport/UdpTransport.php

Tobion: I could the same applies to mongo and elasticsearch handlers as well.

derrabus: While that sounds like a reasonable thing to do, this would change the behavior of the extension. I currently do not have a proper setup to test with Greylog or Elasticsearch, so I don't feel able to perform that change. Maybe someone else could do this in a follow-up PR?

Tobion:
Thank you @derrabus.

PR #250, opened 2018-02-07
Switch to namespaced PHPUnit classes

derrabus:
This PR switches the PHPUnit test base class from PEAR-style `PHPUnit_Framework_TestCase` to the namespaced `PHPUnit\Framework\TestCase`, which enables us to run the tests with PHPUnit 6 as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ matrix:
       env: DEPENDENCIES=dev
     - php: 7.0
     - php: 7.1
+    - php: 7.2
     - php: hhvm
   allow_failures:
     - env: DEPENDENCIES=dev
--- a/Tests/DependencyInjection/Compiler/AddProcessorsPassTest.php
+++ b/Tests/DependencyInjection/Compiler/AddProcessorsPassTest.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Bundle\MonologBundle\Tests\DependencyInjection\Compiler;
 
+use PHPUnit\Framework\TestCase;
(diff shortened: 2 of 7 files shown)


.travis.yml
```diff
@@ -27,7 +27,9 @@ matrix:
     - php: 5.6
       env: DEPENDENCIES=dev
     - php: 7.0
+      env: SYMFONY_PHPUNIT_VERSION=6.5
```
xabbuh: Is this really needed?

derrabus: My intention was to ensure that we maintain compatibility with PHPUnit 6. Without that setting, the PHPUnit bridge would execute PHPUnit 5.7.

https://github.com/symfony/phpunit-bridge/blob/5d27d894072832dce51f73ef3043ce3dfe0f866c/bin/simple-phpunit#L21-L23

xabbuh: What about adding a job for PHP 7.2 instead then?

derrabus: All right, I'll do that. May I pin the php 7.2 job to PHPUnit 6.5? PHPUnit bridge would run PHPUnit 6.3 otherwise.

.travis.yml
```diff
@@ -28,6 +28,8 @@ matrix:
       env: DEPENDENCIES=dev
     - php: 7.0
     - php: 7.1
+    - php: 7.2
+      env: SYMFONY_PHPUNIT_VERSION=6.5
```
stof: no need to specify it. The bridge already handles it in recent versions

derrabus: But maybe this should be fixed in the bridge instead. symfony/symfony#26110

Tobion:
Thank you @derrabus.

PR #251, opened 2018-02-09
inject channel loggers with autowired services

### LINKED ISSUE #218: Autowiring the PSR-3 interface does not correctly map the channels
Configuring the logging channel via a tag does only work if you do not autowire Monolog via a PSR-3 constructor argument. If your service expects a `Psr\Log\LoggerInterface` and you are using autowiring, the `LoggerChannelPass` does not find your definitions. That is IMHO because Symfony autowires the `monolog.logger` service and the compiler pass expects a service argument referring to the `logger` service.

I think the pass should be modified from
```
if ($argument instanceof Reference && 'logger' === (string) $argument) {
```
to
```
if ($argument instanceof Reference && in_array((string) $argument, ['logger', 'monolog.logger'], true)) {
```
(two occurrences)

Then, definitions like the following
```
        <service id="Foo\FooService">
            <argument>logger</argument>
            <argument>%some_param%</argument>
            <tag name="monolog.logger" channel="foo" />
        </service>
```

can simply be written as
```
        <service id="Foo\FooService">
            <argument key="$param">%some_param%</argument>
            <tag name="monolog.logger" channel="foo" />
        </service>
```
even if the expect a PSR-3 `LoggerInterface`.

xabbuh:
This fixes #218.

### DIFF
--- a/DependencyInjection/Compiler/LoggerChannelPass.php
+++ b/DependencyInjection/Compiler/LoggerChannelPass.php
@@ -61,6 +61,38 @@ class LoggerChannelPass implements CompilerPassInterface
                     }
                 }
                 $definition->setMethodCalls($calls);
+
+                if (method_exists($definition, 'setArgument') && $definition->isAutowired() && null !== $class = $definition->getClass()) {
+                    try {
+                        $r = new \ReflectionClass($class);
+                        $constructor = $r->getConstructor();
+
+                        $arguments = $definition->getArguments();
+
+                        foreach ($constructor->getParameters() as $parameter) {
+                            try {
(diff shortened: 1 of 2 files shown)


Tobion:
I guess this does not support autowiring of setters with the channel currently, when used like in the doc https://symfony.com/doc/current/service_container/autowiring.html#autowiring-other-methods-e-g-setters

stof:
A much better solution would be to define a binding for the LoggerInterface on this service, using the right channel (so that autowiring picks the right one), instead of trying to reimplement autowiring.

nicolas-grekas:
Replaced by #254

PR #253, opened 2018-03-01
Enforce that channels defined in the config are public loggers

stof:
Channels can be created for 2 reasons:
- because a service requests to use it through the monolog.logger tag, in which case a private channel is fine (as it is injected using the DIC)
- because the bundle config asks to create it. In this case, the channel should be a public service, as the use case is to be able to use this channel through public access (otherwise, the previous way would have been used).

This is an improved versions over #247 as only channels created through config become public in Symfony 4.

/cc @derrabus


stof:
@lyrixx you were also speaking in favor of keeping the channel public in #247 so this might interest you.

lyrixx:
Nice :+1:

derrabus:
Sounds like a reasonable compromise.

PR #254, opened 2018-03-02
Make channel selection work with autowired arguments

### LINKED ISSUE #218: Autowiring the PSR-3 interface does not correctly map the channels

### LINKED ISSUE #228: Autowiring & tagging by _defaults does not work together
It looks like the MonologBundle has a priority too high, and runs too early, applying channel tagging before that the services have been autowired. 

In my case : 

```yaml
services:
    _defaults:
        autowire: true
        autoconfigure: true
        public: false
        tags:
            - { name: monolog.logger, channel: technical }

    AppBundle\Manager\:
        resource: '../../../src/AppBundle/Manager'
```

My loggers inside my managers are not handled with `technical` channel.

But if i change the service definitions by : 
```yaml
    AppBundle\Manager\:
        resource: '../../../src/AppBundle/Manager'
        arguments:
            - '@logger'
```
it works.

So it looks like MonologBundle find the tag, but there is no logger inside the service to be overridden. Then, the autowiring will run and will put the default logger inside my services, ignoring the tagging.

stof:
This only works on Symfony 3.4+, not on Symfony 3.3, but 3.3 is EOL already so this is fine.

Closes #218 
Closes #228
Replaces #251

### DIFF
--- a/DependencyInjection/Compiler/LoggerChannelPass.php
+++ b/DependencyInjection/Compiler/LoggerChannelPass.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Bundle\MonologBundle\DependencyInjection\Compiler;
 
+use Symfony\Component\DependencyInjection\Argument\BoundArgument;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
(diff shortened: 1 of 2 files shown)


stof:
While discussing with @nicolas-grekas, we realized that the whole logic we have to handle the `logger` references could be replaced by another binding once we drop support for Symfony < 3.4 (which will wait until Symfony 2.8 reaches EOL at the end of the year as we want to keep support for it until then)

PR #257, opened 2018-03-06
require php 5.6

Tobion:
Closes #186. This requires php/symfony versions that are at least supported for security fixes. The doc is outdated with symfony flex and already explained in http://symfony.com/doc/current/logging.html


fabpot:
Thank you @Tobion.

PR #261, opened 2018-04-11
add ident as parameter for syslogudp handler

sandervanmook:
The ident is currently missing as a option for the syslogudp handler.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -138,7 +138,7 @@ use Monolog\Logger;
  *   - [logopts]: defaults to LOG_PID
  *   - [level]: level name or int value, defaults to DEBUG
  *   - [bubble]: bool, defaults to true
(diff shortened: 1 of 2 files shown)


DependencyInjection/Configuration.php
```diff
@@ -138,6 +138,7 @@
  *   - [logopts]: defaults to LOG_PID
  *   - [level]: level name or int value, defaults to DEBUG
  *   - [bubble]: bool, defaults to true
+ *   - [ident]: string, program name or tag for each log message, defaults to php
```
SpacePossum: should be 'php' I think

SpacePossum:
:+1:

Christian-Krupa:
@sandervanmook is there any reason why this issue is still open?

sandervanmook:
The only reason is because it hasn't been merged yet I think ?

lyrixx:
Fixed thanks to #267 

Thanks for working on this feature

PR #268, opened 2018-05-15
Fix the autowiring of the channel logger in autoconfigured services

stof:
This depends on https://github.com/symfony/symfony/pull/27271, hence the new min version of the DI component


fabpot:
Thank you @stof.

PR #270, opened 2018-06-12
Support environment option for raven client

Taluu:
In sentry 9, we can filter logs thourgh environments. This PR allows to support this option.


MatTheCat:
Maybe default value could be `%kernel.environment%`?

Taluu:
I think it's better to leave sentry's default, `null`. It is not really coupled to the kernel's env imo (e.g, on our sentry, we have the `staging`, `qa`, ... and so on)

lyrixx:
Thank you Baptiste.

PR #273, opened 2018-07-25
Register a service.

omrqs:
Register Monolog\Formatter\ElasticaFormatter like a Monolog\Formatter\LogstashFormatter.
Index and type arguments.
Has not public.


lyrixx:
Hello @omrqs 
What is you use case for this ?
Because `index` and `type` will never be the defautl value. So in order to use it, you will need to change the service. So IMHO, this PR is not really useful

stof:
I agree with @lyrixx here. As you would need to override the service anyway to configure it, there is no value into defining it in the bundle.

PR #275, opened 2018-07-26
Deprecrate "excluded_404s" option

geoffrey-brier:
Hello,

Just saw this issue https://github.com/symfony/symfony/issues/26969 and decided to do something about it.

I need to fix the `MonologExtensionTest::testFingersCrossedHandlerWhenExcluded404sAreSpecified` test but I don't really how to proceed, any feedbacks are welcomed :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * Added timeouts to the pushover, hipchat, slack handlers
 * Dropped support for PHP 5.3, 5.4, and HHVM
 * Added configuration for HttpCodeActivationStrategy
+* Deprecated "excluded_404s" option for Symfony >= 3.4
 
 ## 3.2.0 (2018-03-05)
 
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Bundle\MonologBundle\DependencyInjection;
 
+use Symfony\Bridge\Monolog\Handler\FingersCrossed\HttpCodeActivationStrategy;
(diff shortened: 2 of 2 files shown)


stof:
We cannot deprecate it all the time in MonologBundle. We need to detect whether a version of Symfony with the new feature is used. The new way won't be available in the 3.4 LTS at all, while this version of the bundle supports it.

geoffrey-brier:
@stof ok I didn't knew, is this the right place to make a PR then?

stof:
Well, it is. but this PR is incomplete, because the deprecation must be conditional (and the deprecation message also needs to suggest the alternative)

geoffrey-brier:
All right I think I get it, not sure about the Kernel::VERSION_ID though.

stof:
I see 2 issues here:

- you trigger the deprecation for old Symfony version. That's precisely the opposite of what we need: *new* Symfony versions have the new way to exclude errors
- the detection must be based on the component providing the new feature (in a Flex project, the fullstack repo is not used, so you can have different versions of the components). `Kernel::VERSION_ID` will give you the version of `symfony/http-kernel`. the new feature is part of the monolog bridge. Use `class_exists(HttpCodeActivationStrategy::class)` to detect when the new feature is available.

geoffrey-brier:
@stof you're definitely right 👍 Fixed your feedbacks.

Simperfit:
@stof @Seldaek what is missing here ?

lyrixx:
Hello @geoffrey-brier 

I have finished you PR in #302 
Thanks for your work.
Of course, I let you the owernship of the original commit :)

PR #276, opened 2018-08-09
Allow marking service handlers as nested

alcaeus:
Due to the early handling of `service` handlers in the extension, a service handler can't be marked as nested using the `nested` configuration option. This can cause issues when declaring multiple (nested) handlers of which one is then chosen based on the environment: since the service handler can't explicitly be marked as nested, it will always pushed to the logger even when it shouldn't be used.


alcaeus:
Any feedback on this, at all?

Seldaek:
Thanks looks good

PR #277, opened 2018-08-09
Disable process_psr_3_messages for non-leaf handlers

nicolas-grekas:
Currently, logs are processed twice or more. But only leaf handlers should process them by default.

### DIFF
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -142,6 +142,10 @@ class MonologExtension extends Extension
             $definition->setConfigurator(array('Symfony\\Bundle\\MonologBundle\\MonologBundle', 'includeStacktraces'));
         }
 
+        if (null === $handler['process_psr_3_messages']) {
+            $handler['process_psr_3_messages'] = !isset($handler['handler']) && !$handler['members'];
+        }
+
         if ($handler['process_psr_3_messages']) {
             $processorId = 'monolog.processor.psr_log_message';
             if (!$container->hasDefinition($processorId)) {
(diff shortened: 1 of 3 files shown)


DependencyInjection/MonologExtension.php
```diff
@@ -142,6 +142,10 @@ private function buildHandler(ContainerBuilder $container, $name, array $handler
             $definition->setConfigurator(array('Symfony\\Bundle\\MonologBundle\\MonologBundle', 'includeStacktraces'));
         }
 
+        if (null === $handler['process_psr_3_messages']) {
+            $handler['process_psr_3_messages'] = !isset($handler['handler']) && !$handler['members'];
```
lyrixx: This implementation is a bit fragile

nicolas-grekas: Please suggest alternative :)

lyrixx: Play with the container builder to find leaves ?
Or iirc, play with the 'is_nested' flag

nicolas-grekas: But the list of options is finished, there isn't anything more possible than what's listed in Configuration.php, isn't it?
I'm not sure adding a compiler pass to cover something that is basically static is a good idea...

lyrixx: Indeed, "members" and "handler" seems to cover everything.
So basically it works. But if we add something else one day, we will have to remember to update this list. This is why I said it's a **bit** fragile.
But we can live with this ;)

nicolas-grekas: Let's live with this as you say :)

Seldaek:
This whole feature would've been better done by adding the processor at the logger-level and not handler-level IMO, but anyway too late now I suppose so this is still an improvement..

Issue #39, opened 2013-04-12
Add a "monolog:channels" configuration option 


Seldaek:
Defining custom channels for use in controllers right now is impossible since they are created automatically. Having a way to define additional channels from the configuration would be easier to handle. See http://stackoverflow.com/questions/15966140/symfony-2-1-switch-monolog-channel-in-controller

PR #280, opened 2018-09-17
Support the "app" channel in the monolog:channels configuration option

mpdude:
In #39, a configuration option was introduced to manually configure available channels like so:

```yaml
monolog:
    channels: ["foo", "bar"]
```

Before #253, it was also possible to have the `app` channel in this list. This channel corresponds to the `monolog.logger` service.

We can argue about whether it makes sense to declare this channel at all; however, things just worked:tm: because the `LoggerChannelPass::createLogger()` method would just skip over the `app` channel (it's in `LoggerChannelPass::$channels` by default) and nothing else happened to the serivce definition.

Since #253, the definition is turned public which uncovers that not `monolog.logger.app` exists at all. We need to fix this by choosing the right service name for this channel.

***Merge base:*** I don't know about the release policy for this bundle. But you might want to consider this as a bugfix for the v3.2, which does not even require PHP 5.6. But since there is no 3.2 branch in this repo, the PR is against master.


DependencyInjection/Compiler/LoggerChannelPass.php
```diff
@@ -81,7 +81,7 @@ public function process(ContainerBuilder $container)
 
         // create additional channels
         foreach ($container->getParameter('monolog.additional_channels') as $chan) {
-            $loggerId = sprintf('monolog.logger.%s', $chan);
+            $loggerId = $chan === 'app' ? 'monolog.logger' : sprintf('monolog.logger.%s', $chan);
```
stof: this pass should not create the `monolog.logger` service, as it already exists (and the `logger` alias is already public, even though the `monolog.logger` service is not, and was never public)

stof: a better fix would be to skip `app` entirely in this loop, as it is special.

mpdude: If someone declares `app` as a `monolog:channel`, don't you think they'd expect the `monolog.logger` to be `public`?

mpdude:
@stof What about the 3.2 merge base?

DependencyInjection/Compiler/LoggerChannelPass.php
stof: we never made `monolog.logger` public, even before the Symfony change to make services private by default. The public name of the main logger has always been `logger`

stof:
@mpdude we don't maintain multiple branches for minor versions. The next release of the master branch will be either 3.3.1 or 3.4.0 depending on the semver rules (if we only have bug fixes in it, it will be a patch version, otherwise a minor one).

mpdude:
Understand. I‘m not going to admit that this fix would help me if it would be available on PHP 5.5.

DependencyInjection/Compiler/LoggerChannelPass.php
mpdude: And oh, by the way: I just got 

> The "logger" service is private, getting it from the container is deprecated since Symfony 3.2 and will fail in 4.0. You should either make the service public, or stop using the container directly and use dependency injection instead.

... in a logfile. So, what would be the right approach to make `logger` public?

mpdude: @stof *see above*

PR #285, opened 2018-11-10
Register processors for autoconfiguration

nicolas-grekas:
Replaces https://github.com/symfony/symfony/pull/27801 (which has been reverted meanwhile)
Leverages https://github.com/Seldaek/monolog/pull/1204
Already documented in https://github.com/symfony/symfony-docs/pull/9996 (but doc needs update)

### DIFF
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -11,6 +11,9 @@
 
 namespace Symfony\Bundle\MonologBundle\DependencyInjection;
 
+use Monolog\Processor\ProcessorInterface;
+use Symfony\Bridge\Monolog\Processor\TokenProcessor;
+use Symfony\Bridge\Monolog\Processor\WebProcessor;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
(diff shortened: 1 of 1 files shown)


DependencyInjection/MonologExtension.php
```diff
@@ -108,6 +111,18 @@ public function load(array $configs, ContainerBuilder $container)
...
+
+        if (method_exists($container, 'registerForAutoconfiguration')) {
+            if (interface_exists(ProcessorInterface::class)) {
+                $container->registerForAutoconfiguration(ProcessorInterface::class)
+                    ->addTag('monolog.processor');
+            } else {
+                $container->registerForAutoconfiguration(WebProcessor::class)
+                    ->addTag('monolog.processor');
+            }
+            $container->registerForAutoconfiguration(TokenProcessor::class)
```
lyrixx: Why the management of Token is different from WebProcessor?

nicolas-grekas: `WebProcessor extends BaseWebProcessor` while `TokenProcessor` doesn't.

lyrixx:
It took time, but here we go, this is in now. Thank you very much @nicolas-grekas.

PR #289, opened 2018-11-29
Enable the usage of environment variables for log levels.

### LINKED ISSUE #220: It is not possible to use ENV variable for log level

netiul:
This is an solution for #220. It's an alternative to #245 which solves the issue as well but is in my opinion is a bit too rigid/heavy.

This PR modifies the `levelToMonologConst` function to call` \Monolog\Logger::toMonologLevel()` which almost does does the same, but keeps the argument as is without changing it to `NULL` when it does not match a log constant. This way it will be possible to use environment variables as log levels.

For clarity, the \Monolog\Logger::toMonologLevel() functions looks like this:

    public static function toMonologLevel($level)
    {
        if (is_string($level) && defined(__CLASS__.'::'.strtoupper($level))) {
            return constant(__CLASS__.'::'.strtoupper($level));
        }

        return $level;
    }

It will be breaking for those who rely on specifying an invalid level being converted to `NULL`.

Fixes #220

### DIFF
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -32,7 +32,7 @@ class MonologExtension extends Extension
 
     private function levelToMonologConst($level)
     {
-        return is_int($level) ? $level : constant('Monolog\Logger::'.strtoupper($level));
+        return \Monolog\Logger::toMonologLevel($level);
     }
 
     /**
--- a/Tests/DependencyInjection/MonologExtensionTest.php
+++ b/Tests/DependencyInjection/MonologExtensionTest.php
@@ -89,6 +89,16 @@ class MonologExtensionTest extends DependencyInjectionTest
         $this->assertDICConstructorArguments($handler, array('foo', false));
     }
 
+    public function testEnvironmentVariablesAsLogLevel()
+    {
+        $container = $this->getContainer(array(array('handlers' => array('main' => array('type' => 'stream', 'level' => '%env(DUMMY_ENV_LOG_LEVEL)%')))));
+        $this->assertTrue($container->hasDefinition('monolog.logger'));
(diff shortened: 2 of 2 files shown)


Seldaek:
This behavior would break in Monolog v2 as toMonologLevel now throws in case an invalid string is passed.

But given that all Monolog handlers call toMonologLevel anyway on their level arguments, I think the direction of this PR is indeed better than #245 - the bundle doesn't need to make sure the level is resolved, it should just make sure the env var is resolved before being passed in to the handler. /cc @umulmrum I think your PR #245 is nicer for having tests but maybe cleaning it up a little according to the above would be an idea? Unless I missed something..

lyrixx:
@nicolas-grekas Any advice on this one?

umulmrum:
@Seldaek Sorry I didn't find time in the last months to rework my PR. I'm afraid it won't be better in the next weeks or months, so if someone else would like to jump in, I wouldn't oppose ...

lyrixx:
Replaced by #357

Thanks for working on this feature

PR #291, opened 2018-12-14
Remove unused params

phamuyentri:


Seldaek:
Thanks but those parameters are actually passed into the function so I think it's good to have them present for self-documentation for others looking to implement their own factory.

PR #293, opened 2019-01-16
Remove duplicate entry

B-Galati:


lyrixx:
Good catch, thanks @B-Galati.

PR #297, opened 2019-02-22
Expose configuration for the ConsoleHandler

lyrixx:
Like that we will be able to use this configuration (for example):
```yaml
        console:
            type: console
            process_psr_3_messages: false
            channels: ["!event", "!doctrine", "!console"]
            console_formater_options:
                format: "%%datetime%% %%start_tag%%%%level_name%%%%end_tag%% <comment>[%%channel%%]</> %%message%%%%context%%\n"
                multiline: false
```

see also https://github.com/symfony/symfony/pull/30345


OskarStark:
The build is failing, can you please have a look?

lyrixx:
@OskarStark Yes I saw, but it seems not related, isn't ?

PR #300, opened 2019-04-07
Support for Monolog 2

derrabus:
Allows installing Monolog 2 as soon as symfony/symfony#27905 is merged.

### DIFF
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -11,17 +11,20 @@
 
 namespace Symfony\Bundle\MonologBundle\DependencyInjection;
 
+use Monolog\Logger;
 use Monolog\Processor\ProcessorInterface;
 use Monolog\ResettableInterface;
 use Symfony\Bridge\Monolog\Handler\FingersCrossed\HttpCodeActivationStrategy;
 use Symfony\Bridge\Monolog\Processor\TokenProcessor;
 use Symfony\Bridge\Monolog\Processor\WebProcessor;
+use Symfony\Bundle\FullStack;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\HttpKernel\Kernel;
 
 /**
  * MonologExtension is an extension for the Monolog library.
@@ -48,6 +51,10 @@ class MonologExtension extends Extension
      */
     public function load(array $configs, ContainerBuilder $container)
     {
+        if (class_exists(FullStack::class) && Kernel::MAJOR_VERSION < 5 && Logger::API >= 2) {
+            throw new \RuntimeException('Symfony 5 is required for Monolog 2 support. Please downgrade Monolog to version 1.');
+        }
+
         $configuration = $this->getConfiguration($configs, $container);
         $config = $this->processConfiguration($configuration, $configs);
 
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "symfony/dependency-injection": "~3.4.10 || ^4.0.10 || ^5.0",
         "symfony/config": "~3.4 || ~4.0 || ^5.0",
         "symfony/http-kernel": "~3.4 || ~4.0 || ^5.0",
-        "monolog/monolog": "~1.22"
+        "monolog/monolog": "~1.22 || ~2.0"
     },
     "require-dev": {
         "symfony/yaml": "~3.4 || ~4.0 || ^5.0",


Seldaek:
I wonder if this should error if monolog-bridge is in version <5 or not.. Otherwise it might be confusing if Composer lets you update to monolog 2 but then it fails to run.. Maybe worth making a new major release of the bundle for monolog2 + symfony5 support? Thoughts @lyrixx?

Although I guess the monolog-bridge 4.x will prevent installing monolog 2.x and that fixes it? Except for people using symfony/symfony still.

lyrixx:
I'm a bit lost :/ Is this possible to have every version work with every other version?
If not (or too hard), I'm fine with a next major release

Seldaek:
I think it's possible to have symfony5 + monolog1, but not symfony3/4 + monolog2 as the bridge is not going to be made compatible as far as I understand?

derrabus:
> Although I guess the monolog-bridge 4.x will prevent installing monolog 2.x and that fixes it?

Yes, MonologBridge 4.4 and below depends on Monolog 1 and thus will block the installation of Monolog 2.

> Except for people using symfony/symfony still.

That practice is discouraged for over two years now, but yes: Composer won't prevent the installation of Monolog 2 in that case. Anything we can do to detect that case?

derrabus:
> I think it's possible to have symfony5 + monolog1, but not symfony3/4 + monolog2 as the bridge is not going to be made compatible as far as I understand?

If my PR is merged as it is right now, that would be the plan.

Seldaek:
Sounds fine to me to merge as is for now then and avoid a major version of the bundle.

derrabus:
> Except for people using symfony/symfony still.

On second thought: This might still be true for Symfony 3.4 applications. In Symfony 3, it was still best practice to depend on the monorepo.

https://github.com/symfony/symfony-standard/blob/ffb00c045b980f8f721ddf5531c685bfcd79130c/composer.json#L26

I'll try to build some detection logic for that case.

derrabus:
I've added a check for the monorepo case. Composer cannot prevent the installation of `symfony/symfony` 3.4 alongside Monolog 2, but at least we can blow up the container compilation in that case. 🙈

lyrixx:
> I've added a check for the monorepo case. Composer cannot prevent the installation of `symfony/symfony` 3.4 alongside Monolog 2, but at least we can blow up the container compilation in that case.

Why not using the `conflict` directive of composer?

derrabus:
> Why not using the `conflict` directive of composer?

🤔 That could work as well. 

In `symfony/symfony`:
```JSON
"conflict": {
    "monolog/monolog": ">=2"
}
```
If we do this now, Symfony 3.4.30, 4.2.11 and 4.3.3 will have that conflict rule. Then we change composer.json of this PR into
```JSON
"require": {
    "symfony/monolog-bridge": "^3.4.30|~4.2.11|^4.3.3|^5.0"
}
```
and we should be fine. Or am I thinking too complicated? 😅

derrabus:
See symfony/symfony#32469. If that PR gets merged, I will revert my changes to the extension.

Seldaek:
I think I'd merge as is.

inverse:
What's the best way to start working on adding support for new handlers that'll come in the v2 series?

Seldaek:
I think we should first merge this (any objection @lyrixx ? And then could add support for new handlers, but in the Extension class it should check that `Monolog\Logger::API === 2` before allowing using them.

DependencyInjection/MonologExtension.php
```diff
@@ -48,6 +51,10 @@ private function levelToMonologConst($level)
      */
     public function load(array $configs, ContainerBuilder $container)
     {
+        if (class_exists(FullStack::class) && Kernel::MAJOR_VERSION < 5 && Logger::API >= 2) {
```
lyrixx: Why this check on `FullStack` ?

versh23:
monolog 2 has been released

DependencyInjection/MonologExtension.php
derrabus: I want to detect that case that Monolog 2 is installed together with `symfony/symfony` 2.x/3.x/4.x. I don't see a good way to make Composer prevent that combination, so I'm catching that case here. If you know a better way, I'm open for suggestions.

Seldaek: I think the point is the Kernel::MAJOR_VERSION is enough? Or why do you need three datapoints to check two versions?

derrabus: `Kernel::MAJOR_VERSION` is the version of the HttpKernel component, not the version of MonologBridge. But it's the bridge that is actually incompatible before version 5. I have no way to check for the bridge version programmatically. The assumption that the versions of MonologBridge and HttpKernel match is only valid if we have the monorepo.

If I removed the monorepo check, we would also throw if the app uses the standalone packages HttpKernel 4.4, MonologBridge 5.0 and Monolog 2, but that combination should work fine.

Seldaek: Ok gotcha, then sounds good to me to merge as is.. @lyrixx ?

lyrixx: Why do we allow symfony/symfony 4.4 and monolog/monolog 2 in composer, but throw an exception manually? 
I mean, We could forbid installation thanks to composer, isn't ?

Seldaek: That was turned down in https://github.com/symfony/symfony/pull/32469 - IMO this is fine, people using the bundle with the wrong version of symfony+monolog will get an exception, then they can fix.. Not ideal but not so bad either..

Other alternative is to require Monolog 2 and Symfony 5 or at least the monolog bridge in v5 then we fork the bundle completely in two branches for monolog 1 and 2 support.

lyrixx: :+1:

lyrixx (approved):

PR #301, opened 2019-05-02
Fix typo in PhpDoc

jdreesen:
As the `DeduplicationHandler` extends the `BufferHandler` I think the wording for the `handler` parameter should be the same.


lyrixx:
Good catch, thanks Jacob.

PR #292, opened 2019-01-03
Add support to log to redis

ruudvdd:
Updated version of https://github.com/symfony/monolog-bundle/pull/262

PR #305, opened 2019-06-03
Added support for Redis configuration

lyrixx:
fixed https://github.com/symfony/monolog-bundle/pull/292


lyrixx:
It took time, but here we go, this is in now. Thank you very much Ruud Van den Dungen.

PR #306, opened 2019-06-04
Drop support for Symfony <3

lyrixx:


stof:
should we make a release with all the existing changes before that ?

composer.json
```diff
@@ -17,15 +17,15 @@
     ],
     "require": {
         "php": ">=5.6",
-        "symfony/monolog-bridge": "~2.7|~3.3|~4.0",
-        "symfony/dependency-injection": "~2.7|~3.4.10|^4.0.10",
-        "symfony/config": "~2.7|~3.3|~4.0",
-        "symfony/http-kernel": "~2.7|~3.3|~4.0",
+        "symfony/monolog-bridge": "~3.3|~4.0",
```
stof: should we change it to `~3.4` too as 3.3 is EOLed (2.8 still has security support for a few months, but 3.3 does not)

lyrixx: I thought the same thing! But I did not want to bump for the sake of bumping.
But I would go for too

lyrixx:
I don't know. I initially wanted to merge / close as many issue as possible and then make a release.
Before I cleanup this project, the changelog was broken. So making a release without burden could be possible on the very last commit of master f51d377648794e8a8a54f889f17a621911295ce9

But IMHO, it's a bit stupid to do a release, and then to continue to merge new things...

I would go for simplicity: do a release after merge of pending PRs (this one, the one about mailer, and a another couples)

lyrixx:
ping @symfony/deciders

composer.json
fabpot: Let's bump!

fabpot:
I agree, let's do one release with everything, including the bumps for the deps.

Issue #296, opened 2019-02-13
Elasticsearch configuration for host


barell:
Hi all,

At the moment it is not really clear how to configure Elasticearch. I had to spent some time debugging only to find out that when you specify "host" parameter, it can't include http:// part in it. 

I think it would be great to update documentation to mention it, or even better, update code to either accept protocol part of the host URL or throw error so developers know what is going on.

I can create a PR for this but wanted to know your thoughts.

lyrixx:
Since we are using elastica, the simplest thing here is to update the documentation.
Would you mind submit a PR?
Thanks

filkaris:
I took the liberty of opening one with a small update in the documentation

PR #307, opened 2019-06-11
Update Configuration Documentation for Elasticsearch

filkaris:
Related to #296 

Elasticsearch host must not have (http://) or (https://) in the configuration, just the IP or the domain name.

Otherwise an error is thrown that there is `No enabled connection`


lyrixx:
Thank you @filkaris.

PR #313, opened 2019-08-08
#310 : Add sentry 2.0 type

### LINKED ISSUE #310: Support Sentry 2.x with the new Monolog Handler
RavenHandler has been deprecated on Monolog, in favor of Sentry's own Monolog Handler that is shipped with `sentry/sentry`.

So we should support this handler for sentry 2.x.

Taluu:
As said on #310, sentry completely changed their API and even stopped calling their client "raven", using now php-http to handle the requests to their api.

So this PR adds a new type in the config to support the handler they made.

I'm not sure on a few points (such as the self-referencing hub, which I commented because otherwise it would go on an infinite recursion problem), and I based the service definition on [sentry's bundle](https://github.com/getsentry/sentry-symfony).

Poke @Jean85 maybe, as he is maintaining (as far as I can tell ?) the current sentry bundle.

Fixes #310

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## 3.5.0 (xxxx-xx-xx)
 
+* Add `sentry` type to use sentry 2.0 client
+
 ## 3.4.0 (2019-06-20)
 
 * Deprecate "excluded_404s" option
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -668,6 +668,51 @@ class MonologExtension extends Extension
             ));
             break;
 
+        case 'sentry':
+            if (null !== $handler['client_id']) {
+                $clientId = $handler['client_id'];
+            } else {
+                $options = new Definition(
+                    'Sentry\\Options',
+                    array(array('dsn' => $handler['dsn']))
+                );
+
+                if (!empty($handler['environment'])) {
+                    $options->addMethodCall('setEnvironment', [$handler['environment']]);
+                }
+
+                if (!empty($handler['release'])) {
+                    $options->addMethodCall('setRelease', [$handler['release']]);
+                }
+
+                $builder = new Definition('Sentry\\ClientBuilder', array($options));
+
+                $client = new Definition('Sentry\\Client');
+                $client->setFactory(array($builder, 'getClient'));
+
+                $clientId = 'monolog.sentry.client.'.sha1($handler['dsn']);
+                $container->setDefinition($clientId, $client);
+
+                if (!$container->hasAlias('Sentry\\ClientInterface')) {
+                    $container->setAlias('Sentry\\ClientInterface', $clientId);
+                }
+            }
+
+            $hub = new Definition(
+                'Sentry\\State\\Hub',
+                array(new Reference($clientId))
+            );
+
+            // can't set the hub to the current hub, getting into a recursion otherwise...
+            //$hub->addMethodCall('setCurrent', array($hub));
+
+            $definition->setArguments(array(
+                $hub,
+                $handler['level'],
+                $handler['bubble'],
+            ));
+            break;
+
         case 'raven':
             if (null !== $handler['client_id']) {
                 $clientId = $handler['client_id'];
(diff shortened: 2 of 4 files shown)


Taluu:
Removed the check for the class on the config (e.g for raven it was going berserk, as the package is not required to make the tests run). I did the same for sentry 2.0's handler.

DependencyInjection/MonologExtension.php
```diff
@@ -668,6 +668,51 @@ private function buildHandler(ContainerBuilder $container, $name, array $handler
...
+
+            $hub = new Definition(
+                'Sentry\\State\\Hub',
+                array(new Reference($clientId))
+            );
+
+            // can't set the hub to the current hub, getting into a recursion otherwise...
+            //$hub->addMethodCall('setCurrent', array($hub));
+
+            $definition->setarguments(array(
```
lyrixx:
```suggestion
            $definition->setArguments(array(
```

Taluu: FFS, I thought I fixed all of them. Doesn't look like it, but it's now fixed :P

DependencyInjection/MonologExtension.php
```diff
@@ -668,6 +668,51 @@ private function buildHandler(ContainerBuilder $container, $name, array $handler
...
+                    $container->setAlias('Sentry\\ClientInterface', $clientId);
+                }
+            }
+
+            $hub = new Definition(
+                'Sentry\\State\\Hub',
+                array(new Reference($clientId))
+            );
+
+            // can't set the hub to the current hub, getting into a recursion otherwise...
```
Jean85: It's just a singleton. You could use a factory and use the `getCurrent` method to obtain it, but I'm not sure if it would collide with the Sentry Bundle; order of setup of bundles would make a mess here.

DependencyInjection/MonologExtension.php
```diff
@@ -854,6 +899,7 @@ private function getHandlerClassByType($handlerType)
             'socket' => 'Monolog\Handler\SocketHandler',
             'pushover' => 'Monolog\Handler\PushoverHandler',
             'raven' => 'Monolog\Handler\RavenHandler',
+            'sentry' => 'Sentry\Monolog\Handler',
```
Jean85: Raven is the old version, it's completely incompatible with the new version (`sentry/sentry` 2.0).

lyrixx:
> This change as is makes the older Sentry no longer compatible. IDK if you want to proceed like this.

yes it is, 
* raven is for the old one
* sentry is for the new one

Or I missed something?

DependencyInjection/MonologExtension.php
Taluu: That's why I kept the `raven` type, and introduced `sentry` type for the 2.0.

Jean85:
That's correct, but Sentry 2.0 is pretty recent, so Raven\1.x is still pretty widespread.

DependencyInjection/MonologExtension.php
Taluu: As far as I can tell, the singleton part is for the helper functions. I don't think we would want to use these if we're opting for monolog.

With what I made, I think that you can use both (it's just that the hub will be instantiated two times under differents ids, and have the singleton instantiated in the sentry bundle... especially as having both is kinda missing the point IMO).

Taluu:
Then nothing prevents you from using the `raven` while you're still on the 1.0 ?

I could add something in the doc or somewhere else that says that you can use only one of them at a time, not both.

lyrixx:
Could you add a note in the CHANGELOG? 
It will be part of 3.5.0
Thanks

Taluu:
Done

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 ## 3.5.0 (xxxx-xx-xx)
+* Add `sentry` type to use sentry 2.0 client
```
lyrixx: Missing blank line

Taluu: blank line added

lyrixx:
Thanks for your work on this new feature!

PR #314, opened 2019-08-08
Allow Symfony 5

Taluu:
Well, symfony 5 is around the corner, so let's allow it (especially as the master of the skeleton targets 5.0 now)


lyrixx:
Thank you Baptiste.

PR #315, opened 2019-08-10
Symfony autowiring monolog channels #278

### LINKED ISSUE #278: Symfony autowiring monolog channels
Hi everyone,
While looking for a way to autowire monolog channels, and I got stumbled upon StackOverflow thread where guy named "DrafFter" suggests adding LoggerFactory. 
https://stackoverflow.com/questions/43788114/symfony-autowiring-monolog-channels

Maybe this can be implemented within this bundle? Or is there a better way to do autowiring?

adrenalinkin:
resolved #278
- Added possibility for auto-wire monolog channels according to variable type-hint and name.
  Variable will have appropriated name to camel cased monolog channel service name: `monolog.logger.acme -> $acmeLogger`.
- Removed useless import `DefinitionDecorator`.
- Renamed additional channel in the tests from `manualchan` to `additional`

### DIFF
--- a/DependencyInjection/Compiler/LoggerChannelPass.php
+++ b/DependencyInjection/Compiler/LoggerChannelPass.php
@@ -11,13 +11,13 @@
 
 namespace Symfony\Bundle\MonologBundle\DependencyInjection\Compiler;
 
+use Psr\Log\LoggerInterface;
 use Symfony\Component\DependencyInjection\Argument\BoundArgument;
-use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\DependencyInjection\ChildDefinition;
-use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
-use Symfony\Component\DependencyInjection\DefinitionDecorator;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
+use Symfony\Component\DependencyInjection\Reference;
 
 /**
  * Replaces the default logger by another one with its own channel for tagged services.
@@ -28,6 +28,9 @@ class LoggerChannelPass implements CompilerPassInterface
 {
     protected $channels = array('app');
 
+    /**
+     * {@inheritDoc}
+     */
     public function process(ContainerBuilder $container)
     {
         if (!$container->hasDefinition('monolog.logger')) {
@@ -105,11 +108,19 @@ class LoggerChannelPass implements CompilerPassInterface
         }
     }
 
+    /**
+     * @return array
+     */
     public function getChannels()
     {
         return $this->channels;
     }
 
+    /**
+     * @param array $configuration
+     *
+     * @return array
+     */
     protected function processChannels($configuration)
     {
         if (null === $configuration) {
@@ -123,6 +134,13 @@ class LoggerChannelPass implements CompilerPassInterface
         return array_diff($this->channels, $configuration['elements']);
     }
 
+    /**
+     * Create new logger from th monolog.logger_prototype
+     *
+     * @param string $channel
+     * @param string $loggerId
+     * @param ContainerBuilder $container
+     */
     protected function createLogger($channel, $loggerId, ContainerBuilder $container)
     {
         if (!in_array($channel, $this->channels)) {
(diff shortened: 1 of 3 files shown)


adrenalinkin:
@lyrixx Thanks. Now information has been add into CHANGELOG.
Should I add some documentation about new feature?

tristanbes:
I would be interested on how to use that in the README.MD since it adds great DX; (or in the Symfony documentation maybe in the monolog usage ?)

Seldaek:
Yup more docs on this would be good, if I understand correctly you can only name the variable `$monologAcmeLogger` for `monolog.logger.acme` service/`acme` channel?  Any chance of allowing `$acmeLogger`  as well? I feel like the monolog part is kinda superfluous and overly verbose here.

nicolas-grekas:
> acmeLogger

Would make total sense to support only that.
Could be done by passing $channel.'Logger' as 3rd arg to createAliasForArgument.

A full example would be nice I agree because I'm not sure when this would be useful yet :)

adrenalinkin:
@nicolas-grekas @Seldaek Hello. Just my opinion to clarify this decision.

Current logic is pretty simple. A long time documentation teach us how can we find channel service in the DI container by specific logic `monolog.logger.<channel_name>`. 

When auto-wire has been introduce, anyone who previously solves problem with monolog channel auto-wiring, know, how exactly channel services names generates: `acme` become to `monolog.logger.acme`.

So with new feature we are continue this logic - `monolog.logger.acme` become to `Psr\Log\LoggerInterface $monologLoggerAcme`.
Furthermore we get stronger protection from naming collision in the projects.

Seldaek:
IMO naming collision protection isn't needed here as you can only collide with things of class LoggerInterface.. so the monolog name doesn't add anything.

adrenalinkin:
@Seldaek agree, so we can simplify naming according to 
`monolog.logger.acme -> Psr\Log\LoggerInterface $loggerAcme`.

Add the "like" to this comment if this solution acceptable. So I will update code.

About documentation - I am already investigates Symfony Documentation according to all places, where would be good to add information about this new feature. 
I will create PR on this holidays.

Seldaek:
I still find `$acmeLogger` more readable than `$loggerAcme`, If you read like a sentence, you would speak of "the acme logger", or "the logger whose name is acme", but not "the logger acme". 

The service name convention was different, but it doesn't mean we have to carry that over to the variable name.

adrenalinkin:
@Seldaek I am OK to change as you propose.
We just have different principle of the naming.
I was proposed "logger" first because it common for all variables.
`$loggerAcme`, `$loggerDoctrine`, `$loggerSecurity` etc.
Maybe this approach good only for buisness-code not for reusable.

Seldaek:
sorry but I don't get your point, I am only suggesting to swap these, i.e. `$acmeLogger`, `$doctrineLogger`, `$securityLogger`.

adrenalinkin:
@Seldaek no matter, I will change naming as you describes  :)

adrenalinkin:
@Seldaek @nicolas-grekas Hello guys! I was finish with this PR and also create PR in symfony docs.
Just notification about it.

DependencyInjection/Compiler/LoggerChannelPass.php
```diff
@@ -123,6 +134,13 @@ protected function processChannels($configuration)
         return array_diff($this->channels, $configuration['elements']);
     }
 
+    /**
+     * Create new logger from th monolog.logger_prototype
```
stof: typo here. missing `e`

Tests/DependencyInjection/Compiler/LoggerChannelPassTest.php
```diff
@@ -51,7 +51,20 @@ public function testProcess()
             }
         }
 
-        $this->assertNotNull($container->getDefinition('monolog.logger.manualchan'));
+        $this->assertNotNull($container->getDefinition('monolog.logger.additional'));
+
+        // test below acceptable only for Symfony 4.2+
```
stof: make this a separate test, that get skipped

DependencyInjection/Compiler/LoggerChannelPass.php
adrenalinkin: Done

Tests/DependencyInjection/Compiler/LoggerChannelPassTest.php
adrenalinkin: Done

Seldaek:
Thanks

PR #316, opened 2019-08-14
Disable sentry's default integrations when using it

Taluu:
Monolog already does it, as the request integration that comes with the default configuration of sentry.

Disabling it fixes it. Maybe we should add a marker that says we could activate it ? I doubt the interest but I could be wrong.

poke @B-Galati

### DIFF
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -681,7 +681,10 @@ class MonologExtension extends Extension
             } else {
                 $options = new Definition(
                     'Sentry\\Options',
-                    [['dsn' => $handler['dsn']]]
+                    [[
+                        'dsn' => $handler['dsn'],
+                        'default_integrations' => false, // prevent sentry from registering shutdown functions
+                    ]]
                 );
 
                 if (!empty($handler['environment'])) {
@@ -709,9 +712,6 @@ class MonologExtension extends Extension
                 'Sentry\\State\\Hub',
                 [new Reference($clientId)]
             );
-
-            // can't set the hub to the current hub, getting into a recursion otherwise...
-            //$hub->addMethodCall('setCurrent', array($hub));
 
             $definition->setArguments([
                 $hub,


B-Galati (approved):
Great!

B-Galati:
In fact it would be good to keep Request Integration enabled as it gives a lot of useful information.

Taluu:
Thought about it, but as there's already the request processor on monolog, didn't think it'd be useful : https://github.com/Seldaek/monolog/blob/master/src/Monolog/Processor/WebProcessor.php

B-Galati:
That processor is cool but I think it is doing less than Sentry Request Integration.
Also, the goal of the processor is to add information on every single log while Request Integration will add these informations on each captured event which makes more sense for a tool like Sentry where log context is less readable.
Last thing, it's possible that the SentryHandler will not support the "extra" key from log context in future releases. Should be major one but still interesting to know.

Taluu:
> Last thing, it's possible that the SentryHandler will not support the "extra" key from log context in future releases. 

Source ?

Anyway, I managed to readd the request integration. poke @B-Galati @lyrixx

B-Galati:
> > Last thing, it's possible that the SentryHandler will not support the "extra" key from log context in future releases.
> 
> Source ?

https://github.com/getsentry/sentry-python/pull/457#issuecomment-519591414 -> that links to a comment in sentry-php.

B-Galati:
IMHO it would simpler to manage with a factory class that handles all cases.

Like this one: https://github.com/B-Galati/monolog-sentry-handler/blob/master/doc/guide-symfony.md#step-1-configure-sentry-hub

B-Galati:
Also, It is needed to set the global Hub by calling `Hub::setCurrent()`:

```php
// A global HubInterface must be set otherwise some feature provided by the SDK does not work as they rely on this global state
Hub::setCurrent(new Hub($client));
```

lyrixx (approved):
:+1: LGTM.

lyrixx (changes requested):
Do not merge, this is not ready yet.

inverse:
@Taluu needs to be rebased from master due to short array clean up :+1:

Taluu:
Done. I'll try to find a way to set the global hub state, even though it's repulsive (through a factory or a configurator, probably)

jderusse:
Hello what is the status of this PR? why @lyrixx asked for not merging it?

Taluu:
IIRC, it was because sentry relies on a global state, preventing from really having multiple handlers. I can't remember having made any progress, but I think maybe they had some on their side.

We will need to check with sentry's sdk.

lyrixx:
Unfortunately, I can not remember. Sorry.
BTW, je sentry handler that is integrated with monolog is really far from perfect.

In my application, I end up with a full rewrite + a basic "service" handler

B-Galati:
And since then they introduced a new major version.

HypeMC:
Since support for the Sentry type has been deprecated in #537 should we close this one? cc @GromNaN

GromNaN:
Yes.

PR #317, opened 2019-08-16
[RFC] Bump php requirement 

derrabus:
The current master of the bundle aims at becoming compatible with Symfony 5. Although we're still talking about a moving target here, it is very likely that Symfony 5 will add return types to many methods.

A hot candidate would be `ConfigurationInterface::getConfigTreeBuilder()`, a method that this bundle implements. If Symfony 5 adds a return type to that method, this bundle would need to do so as well. That basically means that we would be required to use a syntax element that does not exist in php 5.

I think, this might be the right time to say farewell to php 5 in master. 😃 

This PR bumps the minimal php version to `7.1.3` in order to match the requirements of Symfony 4, but imho we could even be bold here and bump to 7.2.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ vendor
 composer.phar
 composer.lock
 phpunit.xml
+.phpunit.result.cache
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,30 +11,28 @@ cache:
 env:
     global:
         - SYMFONY_PHPUNIT_DIR=.phpunit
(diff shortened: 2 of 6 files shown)


derrabus:
Closed for now as the plans for adding return types have been postponed.

nicolas-grekas:
It would be a good idea to bump here. The plan for Symfony is to encourage third parties to add return types sooner than later, this PR is on the right path!

PR #319, opened 2019-08-27
Improve testing for a specific Symfony version

wouterj:
Using global Symfony Flex helps limiting Symfony versions of the dependencies. Requiring `symfony/symfony` (currently used) means we cannot spot missed dependencies in this bundle for specific versions.

I've also fixed the indentation of the `matrix` section to match the 4 spaces in the rest of the file

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,29 +13,29 @@ env:
         - SYMFONY_PHPUNIT_DIR=.phpunit
 
 matrix:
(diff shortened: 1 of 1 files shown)


wouterj:
Unfortunately, this "improvement" made all tests red. We need flex mostly in the low deps build (PHP 5.6) and there is no Flex version that supports PHP 5.

PR #321, opened 2019-08-31
Supporting v2 handlers and removing removed.

inverse:
This would allow the wiring up of v2 handlers and preventing wiring of removed on V2 API.

Wanted to get feedback if this approach is suitable before I continue with adding tests to cover the validation aspect.

If you like the approach this is taking do you have any guidelines around testing against multiple versions of monolog? e.g. something like this https://github.com/g1a/composer-test-scenarios

Release notes: https://github.com/Seldaek/monolog/releases/tag/2.0.0

### DIFF
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -957,13 +957,6 @@ class MonologExtension extends Extension
             }
         }
 
-        if (Logger::API === 1 && array_key_exists($handlerType, $v2HandlerTypesRemoved)) {
-            @trigger_error(
-                sprintf('"%s" is deprecated and will be removed in MonoLog v2.', $handlerType),
-                E_USER_DEPRECATED
-            );
-        }
-
         if (!isset($typeToClassMapping[$handlerType])) {
             if (Logger::API === 1 && array_key_exists($handlerType, $v2HandlerTypesAdded)) {
                 throw new InvalidArgumentException(
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -947,13 +947,21 @@ class MonologExtension extends Extension
         $v2HandlerTypesRemoved = [
             'hipchat',
             'raven',
+            'slackbot',
         ];
 
         if (Logger::API === 2) {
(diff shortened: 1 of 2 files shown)


inverse:
Thanks @gmponos - added that message. What do you think?

gmponos:
Hello @inverse 

> any experience with multiple version testing? Is that covered by the definitions within travis?

To be honest no. I don't have that much.

>>WDYT?

> What do you think?

Thank you for taking into account my feedback.. I truly wanted also the your opinion (and from the maintainers.. ) if it makes sense to have a deprecation on the configuration level.. It might be overhead since a deprecation will be triggered when the handlers will be initialized:

https://github.com/Seldaek/monolog/blob/1.x/src/Monolog/Handler/HipChatHandler.php#L100

I think that symfony configuration has a special option that you can set a config option as deprecated.. not sure if raises also a deprecation notice and if it would make sense to have it here as well.. 

Last sorry if I caused any trouble...

inverse:
@gmponos Ahh you're right - it probably makes sense to allow the package not the bundle to flag such deprecation. 

As for the Symfony one - looks like that's more for configuration keys which in this case is a string type which I don't think makes sense.

See: https://symfony.com/blog/new-in-symfony-3-4-deprecate-configuration-options

inverse:
@lyrixx any input on this direction?

DependencyInjection/MonologExtension.php
```diff
@@ -938,8 +940,47 @@ private function getHandlerClassByType($handlerType)
             'insightops' => 'Monolog\Handler\InsightOpsHandler',
         ];
 
+        $v2HandlerTypesAdded = [
+            'elasticsearch' => 'Monolog\Handler\ElasticaHandler',
```
inverse: How should this be handed? `ElasticSearchHandler` -> `ElasticaHandler`

but now we have `ElasticsearchHandler` - what should the key be for this one?

lyrixx: Seems good like this to me (elastica != elasticsearch)

fabpot:
Thank you.

PR #324, opened 2019-09-30
add .gitattributes

dmaicher:
Since https://github.com/symfony/symfony/pull/33579 was merged I thought it also makes sense for this bundle.


Seldaek:
Thanks

PR #326, opened 2019-10-07
Add service for MongoDBFormatter

wimme002:
The MongoDBFormatter has sensible values in the constructor so I guess it can be added as a service in this bundle

### DIFF
--- a/Resources/config/monolog.xml
+++ b/Resources/config/monolog.xml
@@ -30,6 +30,7 @@
         <service id="monolog.formatter.json" class="Monolog\Formatter\JsonFormatter" public="false" />
         <service id="monolog.formatter.line" class="Monolog\Formatter\LineFormatter" public="false" />
         <service id="monolog.formatter.loggly" class="Monolog\Formatter\LogglyFormatter" public="false" />
+        <service id="monolog.formatter.mongo_db" class="Monolog\Formatter\MongoDBFormatter" public="false" />
         <service id="monolog.formatter.normalizer" class="Monolog\Formatter\NormalizerFormatter" public="false" />
         <service id="monolog.formatter.scalar" class="Monolog\Formatter\ScalarFormatter" public="false" />
         <service id="monolog.formatter.wildfire" class="Monolog\Formatter\WildfireFormatter" public="false" />


Resources/config/monolog.xml
```diff
@@ -0,0 +1,41 @@
...
+        <service id="monolog.handler.fingers_crossed.error_level_activation_strategy" class="Monolog\Handler\FingersCrossed\ErrorLevelActivationStrategy" abstract="true" />
+
+        <!-- Formatters -->
+        <service id="monolog.formatter.chrome_php" class="Monolog\Formatter\ChromePHPFormatter" public="false" />
+        <service id="monolog.formatter.gelf_message" class="Monolog\Formatter\GelfMessageFormatter" public="false" />
+        <service id="monolog.formatter.html" class="Monolog\Formatter\HtmlFormatter" public="false" />
+        <service id="monolog.formatter.json" class="Monolog\Formatter\JsonFormatter" public="false" />
+        <service id="monolog.formatter.line" class="Monolog\Formatter\LineFormatter" public="false" />
+        <service id="monolog.formatter.loggly" class="Monolog\Formatter\LogglyFormatter" public="false" />
+        <service id="monolog.formatter.mongo_db" class="Monolog\Formatter\MongoDBFormatter" public="false" />
```
jmikola, line 33:
```suggestion
        <service id="monolog.formatter.mongodb" class="Monolog\Formatter\MongoDBFormatter" public="false" />
```

jmikola:
@GromNaN: Am I correct in assuming that defining a service here still requires it to be enabled via the `formatter` option for a handler definition? Per the [custom formatter](https://symfony.com/doc/current/logging/formatter.html) docs.

I expect this eventually would warrant changes to https://github.com/symfony/symfony-docs/blob/7.3/logging/formatter.rst as well to list this in the documentation.

GromNaN:
Looking at https://github.com/Seldaek/monolog/commit/c04d39d10b271b3240f1aae880ff6ff3c964c0c2, the `MongoDBFormatter` is the one by default since Monolog 2.

We can create a new `mongodb` type handler that uses the service `MongoDBFormatter` by default as handler and passes the [constructor options](https://github.com/Seldaek/monolog/blob/2e97231b969e0ffdeff03329b808945b4ba55e38/src/Monolog/Formatter/MongoDBFormatter.php#L34C33-L34C87) (`$maxNestingLevel` and `$exceptionTraceAsString`). Similar to this: https://github.com/symfony/monolog-bundle/blob/43da9a3d05d6d1497adc91137045482f15dd0991/src/DependencyInjection/MonologExtension.php#L668-L678

jmikola:
@GromNaN: I've incorporated the assignment of a default MongoDBFormatter in https://github.com/symfony/monolog-bundle/pull/550

This only applies to handlers created with the new `mongodb` syntax, and the default ctor args of MongoDBFormatter are used. If customer args are required, users can rely on the common `formatter` config syntax and specify a service ID.

GromNaN:
Superseded by https://github.com/symfony/monolog-bundle/pull/550. Thank you @wimme002

PR #329, opened 2019-10-31
[WIP] Configuration Validation

### LINKED ISSUE #325: Validating configuration
We should try to enforce user's correct configuration as commented [here](https://github.com/symfony/monolog-bundle/blob/master/DependencyInjection/Configuration.php#L27-L322).

Why? Because I've found a tricky misconfiguration that leaded to wrong parsing of my logs and in loosing them.

That was the (wrong) configuration stripped off the not congruous parts in order to avoid noise, please don't focus on meaning but keep attention on technical matters

```
monolog:
    channels: ['foo_channel']
    handlers:
        foo_channel_handler:
            channels: [foo_channel]
            type: stream # <--- THE "SILENT ERROR"
            level: error
            handler: grouped
        grouped:
            type:    group
            members: [streamed]
        streamed:
            type:  stream
            path:  "%kernel.logs_dir%/%kernel.environment%.log"
            level: debug
            formatter: monolog.formatter.session_request

monolog.formatter.session_request:
  class: Monolog\Formatter\LineFormatter
  arguments:
    - "[%%datetime%%] [%%extra.token%%] %%channel%%.%%level_name%%: %%message%% \n"
```

This configuration is clearly wrong (`type: stream` tries to use `filter` wrapper handler for a configuration error) but Symfony didn't raise an exception.
Why this could be a problem? 
In this particular case I had two default values explicitated:

`path` --> https://github.com/symfony/monolog-bundle/blob/master/DependencyInjection/Configuration.php#L377
`formatter` (I mean, the Class) --> `Monolog\Formatter\LineFormatter` (AFAIK is the default one)

So in my tests the log was created at the right location using the desired formatter, except that I dind't noticed that `[%%extra.token%%]` was never included. This is due to the fact that, correctly, `stream` type never "pass" control to `grouped` and so to `streamed`.
That was nasty to debug and lost several hours.

We have two chances, both seems to introduce a BC break:

1) Handlers `type` as configuration keys 

Specifying the type of the handler as key and checking that at least one of the the types is specified under handler's name we can better state what to expect and what should not be nested there. One of the things that worry me the most with this approach is we should carry every acceptable value under each `type` key in the configuraton three.
Moreover this would introduce a new way of declaring (and so checking) custom handlers, something like "force" somehow the user to give a definition for its type.

2) Performing semantic checks in `MonologExtension`

This seems the best way from my standpoint. We can easily check [here](https://github.com/symfony/monolog-bundle/blob/master/DependencyInjection/MonologExtension.php#L185-L193) that all expected attributes (and no more that those) are carried by the configuration, raising and error otherwise.

I think that both methods introduce a BC break as in the first case everyone needs to change configuration files whereas in the latter case we were accepting wrong configuration that could possibly not working anymore.

If this makes sense to you I'm willing to work in order to reach what I've shown above.
I'm also open to other ideas.

Let me know what do you think.

DonCallisto:
This fix https://github.com/symfony/monolog-bundle/issues/325

Is worthy to note two main things:

1) I have a failing test (regression?) that I just don't understand why is broken
2) I made all the handler cases specific (ie.: all parameters tested). I know that I can drastically reduce test lines by using a single provider and providing a specific invalid configuration but, in order to reach this, I needo to a) use acceptable parameters from the configuration (for example if I try to use `foo` as "additional" parameter the test will pass but only because `foo` will throw the same exception of "general acceptable" parameter but not accepted for the tested handler), b) if I use accetable parameters as explained in a) I had to make a choice to use all of them or just a random one that's not accepted. I've done all of them but if you prefer to have just a "random one" please let me know and explain why so I can modify it properly.

Thanks for any feedback.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -424,6 +424,7 @@ class Configuration implements ConfigurationInterface
         'logentries' => ['token', 'use_ssl', 'level', 'bubble', 'timeout', 'connection_timeout'],
         'insightops' => ['token', 'region', 'use_ssl', 'level', 'bubble'],
         'flowdock' => ['token', 'source', 'from_email', 'level', 'bubble'],
+        'rollbar' => ['id', 'token', 'config', 'level', 'bubble'],
         'server_log' => ['host', 'level', 'bubble'],
     ];
 
@@ -1069,15 +1070,19 @@ class Configuration implements ConfigurationInterface
             return $v;
         }
 
-        if (!array_key_exists(strtolower($v['type']), $this->acceptedParamsByHandlerType)) {
+// @todo add again, eventually, if we want to raise exception on next if statement
+//        if (in_array($v['type'], ['config', 'id', 'service'])) {
+//            return $v;
+//        }
+
+        if (!array_key_exists($v['type'], $this->acceptedParamsByHandlerType)) {
             return $v;
         }
 
-        // @todo array_keys should be converted to lowercase?
         $acceptableParamsForHandlers = array_intersect($this->provideAllConfigurationParams(), array_keys($v));
         $unacceptableParamForHandler = array_diff(
             $acceptableParamsForHandlers,
-            $this->acceptedParamsByHandlerType[strtolower($v['type'])]
+            $this->acceptedParamsByHandlerType[$v['type']]
         );
         if (!empty($unacceptableParamForHandler)) {
             throw new InvalidConfigurationException(sprintf(
@@ -1093,88 +1098,6 @@ class Configuration implements ConfigurationInterface
 
     private function provideAllConfigurationParams()
     {
-        return [
-            'accepted_levels',
-            'action_level',
-            'activation_strategy',
-            'api_version',
-            'app_name',
-            'auto_log_stacks',
-            'bubble',
-            'bot_name',
-            'buffer_size',
-            'channel',
-            'client_id',
-            'config',
-            'connection_string',
-            'connection_timeout',
-            'console_formater_options',
-            'content_type',
-            'date_format',
-            'deduplication_level',
-            'document_type',
-            'dsn',
-            'elasticsearch',
-            'email_prototype',
-            'environment',
-            'exchange',
-            'exchange_name',
-            'excluded_404s',
-            'excluded_http_codes',
-            'facility',
-            'filename_format',
-            'file_permission',
-            'flush_on_overflow',
-            'from_email',
-            'handler',
-            'headers',
-            'host',
-            'icon_emoji',
-            'id',
-            'ident',
-            'include_extra',
-            'index',
-            'lazy',
-            'level',
-            'logopts',
-            'mailer',
-            'max_files',
-            'max_level',
-            'message_format',
-            'message_type',
-            'min_level',
-            'members',
-            'mongo',
-            'nickname',
-            'notify',
-            'path',
-            'passthru_level',
-            'persistent',
-            'port',
-            'publisher',
-            'redis',
-            'release',
-            'region',
-            'room',
-            'source',
-            'stop_buffering',
-            'store',
-            'subject',
-            'tags',
-            'team',
-            'time',
-            'timeout',
-            'title',
-            'token',
-            'to_email',
-            'url',
-            'user',
-            'use_attachment',
-            'use_locking',
-            'use_short_attachment',
-            'use_ssl',
-            'verbosity_levels',
-            'webhook_url',
-        ];
+        return array_unique(call_user_func_array('array_merge', $this->acceptedParamsByHandlerType));
     }
 }
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -105,7 +105,7 @@ use Monolog\Logger;
  *
  * - fingers_crossed:
  *   - handler: the wrapped handler's name
- *   - [action_level|activation_strategy]: minimum level or service id to activate the handler, defaults to WARNING
+ *   - [action_level|activation_strategy]: minimum level or service id to activate the handler, defaults to null
  *   - [excluded_404s]: if set, the strategy will be changed to one that excludes 404s coming from URLs matching any of those patterns
  *   - [excluded_http_codes]: if set, the strategy will be changed to one that excludes specific HTTP codes (requires Symfony Monolog bridge 4.1+)
  *   - [buffer_size]: defaults to 0 (unlimited)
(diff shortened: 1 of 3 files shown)


DonCallisto:
Moreover, I left this comment as a reminder for myself, maybe also the changelog (and the docs?) should be updated along with this PR.

DonCallisto:
@Seldaek did you had the chance to take a look at my questions and to this PR?

Seldaek:
Yes I had lots of free time so I took a look and then because I am an idiot I decided not to write anything nor to act on that, and just moved on with my life.

DonCallisto:
@Seldaek Sorry to read this.
I was just asking in a polite way seven months after I've been doing this. Was wondering if that was just "backlogged", if it was considered "useless", or whatever; just wish I had an update, nothing more. No prossure with my question.
Take this just as a reminder that you (of course) could have answered in a lot of way, but not harshly as you did.
Sorry for the misunderstanding but, from my standpoint, that's not an acceptable reply.

Seldaek:
From my standpoint it also was not an acceptable intrusion in my inbox.. Hence the harsh reply. I know fully well that there's a ton of issues/PRs rotting away here, it sucks and I hate it but I don't have time. I'm focusing on getting Composer 2 out of the door. This project has other people able to merge stuff so I'm hoping someone else will take care of things.

DonCallisto:
Ok, let me explain.
With my question I didn't intended to put pressure on you.
I wasn't complaining about anything (as you can see).
From time to time (seven months is a "good" temporal span) when I can spend some time on OSS contribution, I make a recap to see if there's something old I was working on and that is still waiting for some reasons. So, since yesterday and for the next days on, I would have been able to work on OSS.
Maybe the question about this test could have been answered in two minutes, maybe not; actually I don't know, so I was politely asking.
My question wasn't a sneaky way to do anything but a question. That's it.
If someone opens a new PR, would you consider it an "invasion" of your inbox?
If someone opens a new issue?
If someone tags you on twitter?
And so on, so on, so on...
What make the difference, here, is the attitude and the insistence of people, I was not harsh at all nor insistent.

How many times I've heard

_"don't open an issue and ask for others to do the work. just contribute yourself"_

Well, it's just what I did. 
After that, knowing that everyone has a life, has other projects, has litte free time, I just waited to know something. And that's part of the game, so it was really natural to me. Also asking politely was natural, considering it was my first question about this.
Let's pretend for a moment I'm a first time OSS contributor overall. Wasn't you over reacting? Wasn't you not welcoming someone, here, with your behavior?

Ok, shit happens and the point is not who is right or who is wrong, I'm just reflecting out loud.
I can imagine how little your free time is, how much effort your putting in composer 2 (that's more "critical" than this library, of course) and all the things, but I was just asking. 
Just a polite question. 

Apart that, I really admire your work and the time you spend on OSS. Composer is great and I'm sure Composer 2 is a big challange. I didn't intended to put pressure on you. I'm honest. If I contribute to OSS is thanks to you, Ocramius, Potencier, etc.
I'm surprised because I saw you to be really polite and that over reaction was a huge shock to me.
But I'm pretty sure you didn't understood my intentions with that question, and, hopefully, now is more clear.

Thanks again for all the effort and the work.

DependencyInjection/Configuration.php
```diff
@@ -963,4 +1062,119 @@ public function getConfigTreeBuilder()
...
+            'to_email',
+            'url',
+            'user',
+            'use_attachment',
+            'use_locking',
+            'use_short_attachment',
+            'use_ssl',
+            'verbosity_levels',
+            'webhook_url',
+        ];
```
Seldaek, line 1178: Could this be generated by recursively merging `$this->acceptedParamsByHandlerType` instead? So it doesn't need to be kept up to date, because I'm pretty sure it'll get outdated otherwise.

DependencyInjection/Configuration.php
```diff
@@ -963,4 +1062,119 @@ public function getConfigTreeBuilder()
...
+    {
+        if (!array_key_exists('type', $v)) {
+            return $v;
+        }
+
+        if (!array_key_exists(strtolower($v['type']), $this->acceptedParamsByHandlerType)) {
+            return $v;
+        }
+
+        // @todo array_keys should be converted to lowercase?
```
Seldaek, line 1076: I don't think so? I don't think there is an expectation that this stuff is case insensitive generally speaking, but I might be mistaken.

DependencyInjection/Configuration.php
```diff
@@ -963,4 +1062,119 @@ public function getConfigTreeBuilder()
...
         return $treeBuilder;
     }
+
+    private function handlerTypeAcceptedOptionsValidation(array $v)
+    {
+        if (!array_key_exists('type', $v)) {
+            return $v;
+        }
+
+        if (!array_key_exists(strtolower($v['type']), $this->acceptedParamsByHandlerType)) {
```
Seldaek, line 1072: Similarly to the note below, strtolower here and at line 1080 probably should be removed.. types were not case insensitive until now?

DependencyInjection/Configuration.php
```diff
@@ -963,4 +1062,119 @@ public function getConfigTreeBuilder()
 
         return $treeBuilder;
     }
+
+    private function handlerTypeAcceptedOptionsValidation(array $v)
+    {
+        if (!array_key_exists('type', $v)) {
+            return $v;
```
Seldaek, line 1069: I think this should rather throw if type is missing? AFAIK there is no valid handler config without a type. Same below if the type is unknown I'd say throw instead of returning a valid value

Seldaek:
Yup sorry for the rant, sometimes it's one too many and someone gets shouted at, not always deservedly so..

Anyway reviewed the PR now as penance.. The test class IMO is way too long and way too much code is spent testing invalid configs which are not really as important as valid ones. Now that the code is there though I don't know if there is much value in removing it.. I just wish it could be done more concisely and perhaps more automated, like generating those invalid config tests using a smarter provider which reads from the available params perhaps having one single provider for valid configs would help there too to be able to array_diff against that in the invalid test provider.  Instead of having to write a test+provider per handler type.

Seldaek:
As for the test failure, I think it's an invalid config in the existing fixtures, https://github.com/symfony/monolog-bundle/blob/master/Tests/DependencyInjection/Fixtures/xml/handlers_with_channels.xml#L33 you should remove the `handler="nested"` part I think that'll fix it.

DonCallisto:
> Yup sorry for the rant, sometimes it's one too many and someone gets shouted at, not always deservedly so..

Nevermind. Really. Sometime happens and realize it is more important that the debate. Don't worry.

> The test class IMO is way too long and way too much code is spent testing invalid configs which are not really as important as valid ones. Now that the code is there though I don't know if there is much value in removing it.. I just wish it could be done more concisely and perhaps more automated, like generating those invalid config tests using a smarter provider which reads from the available params perhaps having one single provider for valid configs would help there too to be able to array_diff against that in the invalid test provider.

Yeah, indeed was one of my biggest worries about this PR. I can think of a more concise way to reach this. Meanwhile I've done the other requested things. Gonna push them in a separate and WIP commit.

Thanks for having reviewed this and thank you for your time.

DependencyInjection/Configuration.php
DonCallisto: By returning `$v`, I let other validations to take place. Here I'm validating only if handler type has correct options, so I think it pretty safe to return `$v` directly. WDYT?

DonCallisto:
I've modified tests (don't pay attention to variable name and functions name, I'll refactor if that's a good approach to you).
One thing is worth to notice: some handlers can manage multiple configurations (like `finger_crossed`). In such cases, the only way to achieve both test's paths (correct configuration one and wrong one) is to keep, only for those handlers that are three (check the test file), old test code.
Otherwise we can remove all wrong configuration tests but I'm not sure this will prevent someone to change the code and mess the things up with validation.

PR #331, opened 2019-11-17
Use an HttpClient without logger in all handlers

lyrixx:
fixes:

* https://github.com/symfony/symfony/pull/34423
* https://github.com/symfony/symfony/pull/32360#issuecomment-554511075
* https://github.com/symfony/symfony-docs/pull/12642

### DIFF
--- a/Resources/config/monolog.xml
+++ b/Resources/config/monolog.xml
@@ -36,5 +36,9 @@
         <service id="monolog.formatter.logstash" class="Monolog\Formatter\LogstashFormatter" public="false">
             <argument index="0">app</argument>
         </service>
+
+        <service id="monolog.http_client" class="Symfony\Contracts\HttpClient\HttpClientInterface" public="false">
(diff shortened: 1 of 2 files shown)


Resources/config/monolog.xml
```diff
@@ -36,5 +36,9 @@
         <service id="monolog.formatter.logstash" class="Monolog\Formatter\LogstashFormatter" public="false">
             <argument index="0">app</argument>
         </service>
+
+        <service id="monolog.http_client" class="Symfony\Contracts\HttpClient\HttpClientInterface">
```
nicolas-grekas: it should be marked as `public="false"` so that the service can be removed

lyrixx: Done. It's only for 3.4, right?

nicolas-grekas: only for 3.4 yes

PR #340, opened 2020-02-28
[codingStandard] factorisation in some method

DurandSacha:
Change some method for light code


Tobion:
Does not seem worth it, esp. since it's only tests

PR #343, opened 2020-03-04
fixes #342 add rfc option for syslogudp handler

### LINKED ISSUE #342: SyslogUdpHandler - Allow setting "rfc" parameter
The SyslogUdpHandler constructor has a last parameter "rfc", default value is ["self::RFC5424"](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/SyslogUdpHandler.php#L47).
However, the current Monolog Bundle implementation dos not allows to set another value like self::RFC3164.

It will be nice if we can do something like that:

```
monolog:
    handlers:
        syslog:
            type: syslogudp
            host: 127.0.0.1
            port: 514
            level: error
            ident: app
            rfc: 0
```

gwinn:


lyrixx (approved):

DependencyInjection/Configuration.php
```diff
@@ -477,6 +478,7 @@ public function getConfigTreeBuilder()
                             ->scalarNode('title')->defaultNull()->end() // pushover
                             ->scalarNode('host')->defaultNull()->end() // syslogudp & hipchat
                             ->scalarNode('port')->defaultValue(514)->end() // syslogudp
+                            ->scalarNode('rfc')->defaultNull()->end() // syslogudp
```
Tobion: should be an integerNode as SyslogUdpHandler constructor requires an int

Tobion: also please add description and potential values like \Monolog\Handler\SyslogUdpHandler::RFC3164

DependencyInjection/Configuration.php
```diff
@@ -157,6 +157,7 @@
  *   - [level]: level name or int value, defaults to DEBUG
  *   - [bubble]: bool, defaults to true
  *   - [ident]: string, defaults to
+ *   - [rfc]: string, defaults to
```
Tobion: defaults to ...? it does not have a default. also it's an integer

Tobion (changes requested):

DependencyInjection/Configuration.php
gwinn: It has, https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/SyslogUdpHandler.php#L47

But yes, it is not correct to describe it as string

gwinn:
Can you review latest changes?

lyrixx (approved):
:+1:

DependencyInjection/Configuration.php
```diff
@@ -477,6 +478,7 @@ public function getConfigTreeBuilder()
                             ->scalarNode('title')->defaultNull()->end() // pushover
                             ->scalarNode('host')->defaultNull()->end() // syslogudp & hipchat
                             ->scalarNode('port')->defaultValue(514)->end() // syslogudp
+                            ->integerNode('rfc')->defaultValue(1)->end() // syslogudp
```
Tobion, line 481: I think this should be an enumNode and list the possible values. Also please use the constants from \Monolog\Handler\SyslogUdpHandler instead of plain int.

DependencyInjection/Configuration.php
```diff
@@ -157,6 +157,7 @@
  *   - [level]: level name or int value, defaults to DEBUG
  *   - [bubble]: bool, defaults to true
  *   - [ident]: string, defaults to
+ *   - [rfc]: int 0 (RFC3164) or 1 (RFC5424), defaults to 1
```
Tobion, line 160: please list the constants from \Monolog\Handler\SyslogUdpHandler instead of plain int

Tobion (changes requested):

iyzoer:
Hi! When can we expect a merge? This very important for me. It seems like nothing else prevents the merge.

HontoNoRoger:
Also highly requested from my side, need it to select RFC5424e (time with miliseconds) for SyslogUDP logging.

dbrumann (commented):
It would be nice to have a test case in Tests/DependencyInjection/ConfigurationTest.php to see if the default is used when rfc is not set in config and another case for when an invalid number is picked.

DependencyInjection/Configuration.php
```diff
@@ -486,6 +488,10 @@ public function getConfigTreeBuilder()
                             ->scalarNode('title')->defaultNull()->end() // pushover
                             ->scalarNode('host')->defaultNull()->end() // syslogudp & hipchat
                             ->scalarNode('port')->defaultValue(514)->end() // syslogudp
+                            ->enumNode('rfc')
+                                ->values([SyslogUdpHandler::RFC5424, SyslogUdpHandler::RFC3164])
```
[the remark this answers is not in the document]
gwinn: That's why i'm not sure about this RFC at this moment.
Should it be added at UdpHandler in Monolog 1? Or should it be added in some other place and imported alongside with SuslogUdpHandler for Monolog v1.x?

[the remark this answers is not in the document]
dbrumann: I am not sure how actively maintained v1 is, i.e. whether new features are accepted and if this is considered a new feature.

Both UdpHandler [v1](https://github.com/Seldaek/monolog/blob/1.x/src/Monolog/Handler/SyslogUdpHandler.php#L98-L109) and [v2](https://github.com/Seldaek/monolog/blob/main/src/Monolog/Handler/SyslogUdpHandler.php#L104-L115) will just ignore invalid values as far as I can tell. I think we can just allow users to put in whatever int they want, maybe add a comment for where to find the proper values (e.g. "see UdpHandler RFC-constants") and not bother with the constants at all. Not sure what others think about this, though.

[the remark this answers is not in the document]
gwinn: Well, i think we should leave this story as is at this moment, we can add support for RFC5424e into another PR to speed up the acceptance of the current request

dbrumann (approved):

DependencyInjection/Configuration.php
gwinn: Got it. In any case, the user should understand that monolog.yaml should look like

```
monolog:
    handlers:
        syslog:
            type: syslogudp
            host: 127.0.0.1
            port: 514
            level: error
            ident: app
            rfc: 0
```

gwinn:
> It would be nice to have a test case in Tests/DependencyInjection/ConfigurationTest.php to see if the default is used when rfc is not set in config and another case for when an invalid number is picked.

Done

DependencyInjection/Configuration.php
```diff
@@ -486,6 +488,10 @@ public function getConfigTreeBuilder()
...
+                                ->defaultValue(SyslogUdpHandler::RFC5424)
+                                ->end() // syslogudp
```
jderusse, line 494: remove extra indentaion

DependencyInjection/MonologExtension.php
```diff
@@ -520,6 +520,9 @@ private function buildHandler(ContainerBuilder $container, $name, array $handler
             if ($handler['ident']) {
                 $definition->addArgument($handler['ident']);
             }
+            if (isset($handler['rfc'])) {
```
jderusse, line 523: Hmm what if `ident` is null/false and this value is set, I believe, you'll end up with `$handler['rfc']` injected in `$ident` argument.
Could you add a test to assert that scenario won't happend?

jderusse (changes requested):

DependencyInjection/MonologExtension.php
gwinn: Seems like everything is ok

gwinn:
Can I expect a review soon? It seems to me important and necessary to contribute to open source projects, especially if the proposed changes are needed not only by one developer, but, frankly, it is a little confusing how quickly the library development team considers proposals, this PR was created a year ago

DependencyInjection/Configuration.php
jderusse, line 492: using raw value (cf snippet bellow) would allow people to use the RFC, without triggering exception for composer 1

```suggestion
                                ->values([SyslogUdpHandler::RFC5424, SyslogUdpHandler::RFC3164, 2 /* SyslogUdpHandler::RFC5424e */])
```

DependencyInjection/MonologExtension.php
jderusse: nop.
1. it does not work
2. you didn't added test for it

```
class MonologExtensionTest extends DependencyInjectionTest
{
    public function testSyslog()
    {
        $container = new ContainerBuilder();

        $loader = new MonologExtension();
        $configs = [
            [
                'handlers' => [
                    'syslogudp' => [
                        'type' => 'syslogudp',
                        'host' => '127.0.0.1',
                        'port' => 514,
                        'facility' => 'USER',
                        'ident' => null,
                        'level' => 'ERROR',
                        'rfc' => SyslogUdpHandler::RFC5424,
                    ]
                ]
            ]
        ];
        $expectedArguments = [
            '127.0.0.1', // $host
            514, // $port
            'USER', // $facility
            400, // $level
            true, // $bubble
            'php', // $ident,
            1, // $rfc
        ];
        $loader->load($configs, $container);

        $definition = $container->getDefinition('monolog.handler.syslogudp');
        $this->assertDICConstructorArguments($definition, $expectedArguments);
    }
}
```

> 1) Symfony\Bundle\MonologBundle\Tests\DependencyInjection\MonologExtensionTest::testSyslog
> Expected and actual DIC Service constructor arguments of definition 'Monolog\Handler\SyslogUdpHandler' don't match.
> Failed asserting that two arrays are equal.
```diff
--- Expected
+++ Actual
@@ @@
     2 => 'USER'
     3 => 400
     4 => true
-    5 => 'php'
-    6 => 1
+    5 => 1
 )
```

jderusse (changes requested):

DependencyInjection/MonologExtension.php
dbrumann: good catch

DependencyInjection/Configuration.php
gwinn: https://github.com/symfony/monolog-bundle/pull/343#discussion_r470972897 here @Tobion suggest to not to use plain values, which point of view should I adhere to

DependencyInjection/MonologExtension.php
gwinn: Yep, my bad. 

https://github.com/Seldaek/monolog/blob/main/src/Monolog/Handler/SyslogUdpHandler.php#L49 here i see that ident has default value and it's a string

Maybe we should make this argument required with the same default value as into SyslogUdpHandler?

DependencyInjection/Configuration.php
jderusse: It's always better to use constant, that's why @Tobion suggested to not use plain value when a constant exists.

In this case (and only for the value `2`), we can not use constant, as the constant does not exist in monolog 1.
But we should not prevent people that use monolog 2 to use this value.
As a result, you should either remove this ENUM constraint and let people enter whatever they want, or allows this rfc by using a plain value.

gwinn: Ok, i'll change it to scalarNode with default value

DependencyInjection/MonologExtension.php
gwinn: I mean that we can change this `->scalarNode('ident')->defaultFalse()->end() // syslog and syslogudp` to `->scalarNode('ident')->defaultValue('php')->end() // syslog and syslogudp` and change this

```

Comment (jderusse) [reply DependencyInjection/MonologExtension.php:525]: IMHO `->scalarNode('ident')->defaultValue('php')->end()` is simpler

Comment (gwinn) [reply DependencyInjection/MonologExtension.php:525]: yep, it's simpler, but if we only change to `->scalarNode('ident')->defaultValue('php')->end()` than [problem](https://github.com/symfony/monolog-bundle/pull/343#discussion_r601499445) still happens

File: DependencyInjection/MonologExtension.php [hunk 8]
```diff
@@ -519,12 +519,18 @@ private function buildHandler(ContainerBuilder $container, $name, array $handler
...
                 $handler['bubble'],
             ]);
-            if ($handler['ident']) {
+
+            if (false === $handler['ident'] || null === $handler['ident']) {
+                $handler['ident'] = 'php';
+                $definition->addArgument($handler['ident']);
+            } elseif ($handler['ident']) {
                 $definition->addArgument($handler['ident']);
             }
```
jderusse, line 528: does not work with other "false" values like `0`, `''`, ...
I suggest to always set a value, and use an `if/else` block for that
```suggestion
            if ($handler['ident']) {
                $definition->addArgument($handler['ident']);
            } else {
                $handler['ident'] = 'php';
                $definition->addArgument($handler['ident']);
            }
```

jderusse (commented):
A last comment for my side ;-).

Don't forget to add an entry in the ChangeLog :pray:

gwinn:
> A last comment for my side ;-).
> 
> Don't forget to add an entry in the ChangeLog pray

done )

jderusse (approved):

DependencyInjection/Configuration.php
```diff
@@ -408,7 +410,7 @@ public function getConfigTreeBuilder()
                             ->booleanNode('use_locking')->defaultFalse()->end() // stream and rotating
                             ->scalarNode('filename_format')->defaultValue('{filename}-{date}')->end() //rotating
                             ->scalarNode('date_format')->defaultValue('Y-m-d')->end() //rotating
-                            ->scalarNode('ident')->defaultFalse()->end() // syslog and syslogudp
+                            ->scalarNode('ident')->defaultValue('php')->end() // syslog and syslogudp
```
Tobion, line 413: This looks like a bc break. The https://github.com/Seldaek/monolog/blob/main/src/Monolog/Handler/SyslogHandler.php does not have 'php' as default value. So it used an empty string (casted from false) before.

gwinn: But we talking about SyslogUdpHandler https://github.com/Seldaek/monolog/blob/main/src/Monolog/Handler/SyslogUdpHandler.php#L49

jderusse: yes, but @Tobion is right here, this configuration is also used by both `Syslog` and `SyslogUdp`.
The change should keep compatibility with the 2 handlers.

Could you also please add a test to cover this case?

GromNaN:
Rebased onto `4.x` and adapted to the current codebase (`src/` layout, modern Monolog API).

- Add `rfc` config option for the `syslogudp` handler (defaults to `RFC5424`, accepts `RFC3164`/`RFC5424e`).
- Pass `ident` and `rfc` explicitly to `SyslogUdpHandler`, falling back to `'php'` when `ident` is falsy (the handler now requires a string).
- Change the `ident` config default from `false` to `'php'`, which also fixes a latent `TypeError` on the `syslog` handler.
- Add the `rfc` attribute to the XML schema and cover the new option with tests.

GromNaN (approved):

GromNaN:
Thank you @gwinn.

Issue #345, opened 2020-03-16
The child node "type" at path "monolog.handlers.redis" must be configured.


shahariaazam:
Getting error `The child node "type" at path "monolog.handlers.redis" must be configured.`.

As per the sample configuration for `redis` (found from https://github.com/symfony/monolog-bundle/blob/master/DependencyInjection/Configuration.php#L91)

Here is my configuration.

```
 redis:
    redis:
        #id: optional if host is given
        host: 127.0.0.1
        password: null
        port: 6379
        database: 0
        key_name: monolog_redis
```

But it's throwing exception "The child node "type" at path "monolog.handlers.redis" must be configured.".

I think either I am missing something or that sample configuration needs to be updated

shahariaazam:
Is anyone here to review this PR?

jderusse:
The docbloc you are pointing is not a sample configuration, it is the list of parameters for each `type` of handler.

You'll find sample configuration of handlers is the documentation: https://symfony.com/doc/current/logging.html

PR #346, opened 2020-03-16
Issue #345 fixed

shahariaazam:
Just updated the reference by adding `type` under `monolog.handlers.redis`

```
        redis:
            type: redis
            redis:
                #id: optional if host is given
                host: 127.0.0.1
                password: null
                port: 6379
                database: 0
                key_name: monolog_redis
```

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -89,6 +89,7 @@ use Monolog\Logger;
  *   - [bubble]: bool, defaults to true
  *
  * - redis:
+ *   - type: redis
  *   - redis:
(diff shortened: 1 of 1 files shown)


jderusse:
The `type` attribute have to be configured whatever the value inside the configuration (this is not just about `redis`, but all the handler types)

The purpose of this docbloc is to list the different options for each `type`. See comment on top of the docbloc you are updating https://github.com/symfony/monolog-bundle/blob/a9122a26bcc08ba11a8c9d8cbc6ce168adff1e86/DependencyInjection/Configuration.php#L25

PR #349, opened 2020-04-28
Allow to inject multiple loggers

l-vo:
To use multiple channels in the same class when autowire isn't enabled

### DIFF
--- a/DependencyInjection/Compiler/LoggerChannelPass.php
+++ b/DependencyInjection/Compiler/LoggerChannelPass.php
@@ -39,6 +39,7 @@ class LoggerChannelPass implements CompilerPassInterface
 
         // create channels necessary for the handlers
         foreach ($container->findTaggedServiceIds('monolog.logger') as $id => $tags) {
+            $channelsForTags = [];
             foreach ($tags as $tag) {
                 if (empty($tag['channel']) || 'app' === $tag['channel']) {
                     continue;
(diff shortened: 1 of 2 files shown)


stof:
Relying on the order of tags vs the order in which arguments referencing `logger` are discovered looks very fragile to me.

l-vo:
Yes, it was a fragile implementation, but since it's an edge case (due to the rise of autowiring), I didn't want to add too much complexity. I've just pushed another proposition with a better balance (IMHO) between solidity and complexity.

stof:
Well, if you configure arguments manually, you can already reference `monolog.logger.foo` and `monolog.logger.bar` directly to inject the channels you want.

l-vo:
Yes of course, but in this case the app is highly tied with how the logger services id are built. Isn't it one of the purposes of the `LoggerChannelsPass` to avoid a too strong relation between MonologBundle and the application (or the framework) ?

stof:
well, the goal of the `monolog.channel` tag with a generic `logger` reference was indeed to reduce the coupling to MonologBundle. But that was mostly for shared bundles than for apps (apps will be the one making a choice of which logging bundle they use anyway).

But here, this PR introduce more complexity with a fragile implementation to handle an edge case. IMO, it is fine to have coupling in this case.

l-vo:
It's not really for an app but for an upcoming PR on core. But I'm ok it's a very edge case. So I'm going to use direct injection in my case. Thank you for your feedback 👌

PR #350, opened 2020-05-07
Add container.preload tag on all handlers

fancyweb:
The `ConsoleHandler` is an event subscriber and therefore gets the `container.no_preload` tag automatically. However, we want to preload this class because it is in the hot path. So we have to add the `container.preload` tag to it (see https://github.com/symfony/symfony/pull/36742).

Instead of just adding the tag for this handler, it's better to add it for all (automatically) because all handlers are always pushed to a logger.


nicolas-grekas:
> However, we want to preload this class because it is in the hot path

this sentence made me realize this might be a better option to fix the issue:
https://github.com/symfony/symfony/pull/36749

If you agree, this means we could save us from merging this PR in here.

PR #351, opened 2020-05-21
fix tests for php 7.x

mipxtx:
unit test are broken for 7+ versions at master. This fixes it

### DIFF
--- a/Tests/DependencyInjection/Fixtures/yml/multiple_email_recipients.yml
+++ b/Tests/DependencyInjection/Fixtures/yml/multiple_email_recipients.yml
@@ -6,3 +6,6 @@ monolog:
             to_email:   [dev1@example.com, dev2@example.com]
             subject:    An Error Occurred!
             level:      debug
+services:
+    mailer:
+        class:     Mailer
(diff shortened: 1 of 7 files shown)


Tests/DependencyInjection/Fixtures/yml/multiple_email_recipients.yml
```diff
@@ -6,3 +6,6 @@ monolog:
             to_email:   [dev1@example.com, dev2@example.com]
             subject:    An Error Occurred!
             level:      debug
+services:
+    mailer:
+        class:     Mailer
```
Tobion, line 11: missing new line

Tests/DependencyInjection/MonologExtensionTest.php
```diff
@@ -538,7 +556,7 @@ public function testFingersCrossedHandlerWhenExcludedHttpCodesAreSpecified()
      * @param string $handlerType
```
Tobion, line 556: this param name and type seems wrong now

composer.json
```diff
@@ -26,7 +26,8 @@
     "require-dev": {
         "symfony/yaml": "~3.4 || ~4.0 || ^5.0",
         "symfony/console": "~3.4 || ~4.0 || ^5.0",
-        "symfony/phpunit-bridge": "^3.4.19 || ^4.0 || ^5.0"
+        "symfony/phpunit-bridge": "^3.4.19 || ^4.0 || ^5.0",
+        "phpunit/phpunit": "<8.0"
```
Tobion, line 30: why require phpunit? it makes use of symfony/phpunit-bridge which installs phpunit itself in a version compatible with php

PR #353, opened 2020-06-04
Option to set order of processors

Erikvv:
Currently you cannot determine the order in which processors are added, and thus called.

In other parts of symfony this is done through a "priority" attribute.

I have added this and I'm using this in a proprietary project.

If it is accepted I will also make a PR for the documentation

### DIFF
--- a/DependencyInjection/Compiler/AddProcessorsPass.php
+++ b/DependencyInjection/Compiler/AddProcessorsPass.php
@@ -28,32 +28,57 @@ class AddProcessorsPass implements CompilerPassInterface
             return;
         }
 
+        $processors = [];
+
         foreach ($container->findTaggedServiceIds('monolog.processor') as $id => $tags) {
             foreach ($tags as $tag) {
-                if (!empty($tag['channel']) && !empty($tag['handler'])) {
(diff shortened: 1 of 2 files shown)


DependencyInjection/Compiler/AddProcessorsPass.php
```diff
@@ -28,32 +28,57 @@ public function process(ContainerBuilder $container)
...
+            $processors,
+            function (array $left, array $right) {
+               return $left['tag']['priority'] - $right['tag']['priority'];
+            }
+        );
+
+        foreach ($processors as $processor) {
+            $tag = $processor['tag'];
+            $id = $processor['id'];
+
```
Erikvv, line 58: Below this line the code is identical to what it was, except 1 level less indented

natepage:
Any news on this PR, we have a project that requires this feature. Anything we can do to help make it move forward?

jderusse (changes requested):
Symfony exposes a `PriorityTaggedServiceTrait` trait that provides all the logic you need in this PR.

This would also deserve an entry in the CHANGELOG

Erikvv:
I am no longer in the ecosystem. Someone else can make a new PR.

PR #357, opened 2020-09-09
Set log level from parameter or envvar.

dbrumann:
This PR is yet another attempt at solving issue #220, just like #245 and #289 before.

Sorry, for opening yet another PR for this, but since there wasn't much activity I figured I give it a go.

### DIFF
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -23,6 +23,7 @@ use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\Argument\BoundArgument;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Exception\ParameterNotFoundException;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
@@ -41,9 +42,31 @@ class MonologExtension extends Extension
 
     private $swiftMailerHandlers = [];
 
-    private function levelToMonologConst($level)
+    private function levelToMonologConst($level, ContainerBuilder $container)
     {
-        return is_int($level) ? $level : constant('Monolog\Logger::'.strtoupper($level));
+        if (null === $level || is_numeric($level)) {
+            return $level;
+        }
+
+        if (defined('Monolog\Logger::'.strtoupper($level))) {
+            return constant('Monolog\Logger::' . strtoupper($level));
+        }
+
+        if ($container->hasParameter($level)) {
+            return $this->levelToMonologConst($container->getParameter($level), $container);
+        }
+
+        try {
+            $logLevel = $container->resolveEnvPlaceholders($level, true);
+        } catch (ParameterNotFoundException $notFoundException) {
+            throw new \InvalidArgumentException(sprintf('Could not match "%s" to a log level.', $level));
+        }
+
+        if ($logLevel !== '') {
+            return $this->levelToMonologConst($logLevel, $container);
+        }
+
+        throw new \InvalidArgumentException(sprintf('Could not match "%s" to a log level.', $level));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


SerkanYildiz:
Hi @dbrumann, we're in the phase of extending the Extension class of Monolog to have this functionality.
However we would like to have this in the bundle, so that we don't manage it ourselves.

Can we help you to get this functionality merged in the bundle itself?

Thanks in advance!

dbrumann:
Thanks @SerkanYildiz. The failing tests should be addressed by #360. I hope after merging that we can also merge this. I guess other than reviewing if this is the appropriate way to solve this, there is not much else to do here, right now.

DependencyInjection/MonologExtension.php
```diff
@@ -41,9 +42,31 @@ class MonologExtension extends Extension
 
     private $swiftMailerHandlers = [];
 
-    private function levelToMonologConst($level)
+    private function levelToMonologConst($level, ContainerBuilder $container)
     {
-        return is_int($level) ? $level : constant('Monolog\Logger::'.strtoupper($level));
+        if ($level === null || is_int($level)) {
```
lyrixx, line 47: Why do you allow returning  `null`?

Tests/DependencyInjection/MonologExtensionTest.php
```diff
@@ -611,10 +612,83 @@ public function v1AddedDataProvider()
...
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('Could not match "%some_param%" to a log level.');
+
+        $loader->load($config, $container);
+    }
+
+    /**
+     * @dataProvider provideLoglevelParameterConfig
+     */
+    public function testLogLevelfromParameter($parameters, $config, $expectedClass, $expectedArgs)
```
lyrixx, line 630:
```suggestion
    public function testLogLevelfromParameter(array $parameters, array $config, string $expectedClass, array $expectedArgs)
```

Tests/DependencyInjection/MonologExtensionTest.php
```diff
@@ -611,10 +612,83 @@ public function v1AddedDataProvider()
...
+        foreach ($parameters as $name => $value) {
+            $container->setParameter($name, $value);
+        }
+        $loader = new MonologExtension();
+        $config = [['handlers' => ['main' => $config]]];
+        $loader->load($config, $container);
+
+        $definition = $container->getDefinition('monolog.handler.main');
+        $this->assertDICDefinitionClass($definition, $expectedClass);
+
```
lyrixx, line 642:
```suggestion
```

lyrixx (commented):
Looks good, but I left some comments. 
Thanks for working on this

DependencyInjection/MonologExtension.php
dbrumann: I figured not specifying a level will not pass it and cause the handler's default to be used. I will revert it

lyrixx:
Could you add a note in the CHANGELOG? after that I'm 👍🏼

lyrixx (approved):

dbrumann:
I just tried this out in a project and noticed an issue. The following config (when no env var is set) will not work:

```yaml
# config/packages/monolog/dev.yaml
parameters:
    env(LOG_LEVEL): ~
monolog:
    handlers:
        main:
            type: stream
            path: "%kernel.logs_dir%/%kernel.environment%.log"
            level: '%env(LOG_LEVEL)%'
```

I am working on a fix for this.

lyrixx (approved):
OK, let's go 👍🏼

lyrixx:
It took time, but here we go, this is in now. Thank you very much @dbrumann.

eexit:
Hello,

Quick question: is it possible to combine this env var with the default resolver?
I'm trying to do this but I'm getting the following error: `'%env(default:info:LOG_LEVEL)%'`.

```
In MergeExtensionConfigurationPass.php line 217:
!!
!!    Using a cast in "env(default:info:LOG_LEVEL)" is incompatible with resoluti
!!    on at compile time in "Symfony\Bundle\MonologBundle\DependencyInjection\Mon
!!    ologExtension". The logic in the extension should be moved to a compiler pa
!!    ss, or an env parameter with no cast should be used instead.
```

Any insight?
Thanks!

dbrumann:
No, unfortunately not.  This ties into the issue reported in #413 The log level needs to be evaluated during container compilation to be able to build the logger.

As far as I can tell we can only get the value and not use resolvers at that point. I am not 100% sure if this is really a general limitation or caused by how I implemented it. It might also have changed - I haven't really kept up to date with the changes around resolving env vars. So it might be doable, but I am afraid I don't have much time to check it at the moment. Feel free to have a deeper look, if you can.

eexit:
Ok, thanks @dbrumann!

PR #359, opened 2020-09-18
Add PHP 7.4 to Travis test matrix

dbrumann:

PR #358, opened 2020-09-18
Adds ForwardCompatTestTrait and fixes related Test

dbrumann:
Fixes PHP Fatal error in newer PHP versions:

> PHP Fatal error:  Declaration of  Symfony\Bundle\MonologBundle\Tests\DependencyInjection\Compiler\AddSwiftMailerTransportPassTest::setUp() must be compatible with PHPUnit\Framework\TestCase::setUp(): void in /Volumes/Projects/symfony/monolog-bundle/Tests/DependencyInjection/Compiler/AddSwiftMailerTransportPassTest.php on line 29

I copied the ForwardCompatTestTrait from Symfony\Validator: https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Validator/Test/ForwardCompatTestTrait.php

PR #360, opened 2020-09-18
Fix tests for php7

dbrumann:
Combines previous PRs #358 and #359 together with fixes for failing tests on Travis.

 * Updates travis test matrix:
   * Adds PHP 7.4 to test matrix
   * Expands test matrix for PHP 7.2+ to check against both monolog 1 & 2
 * Fixes PHPUnit deprecations:
   * Fixes invalid method signature for `setUp()`
   * Replaces deprecated assertions (assertContains, assertNotContains)
   * Replaces deprecated expectedExceptions annotations
 * Fixes missing service definitions in container configurations

Closes #351 as well

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,17 +24,32 @@ matrix:
     - php: 7.1
       # There is a bug in PHPUnit 5.7
       env: SYMFONY_PHPUNIT_VERSION=6.5
+    # Test against seldaek/monolog 1.x
     - php: 7.2
+      env: MONOLOG_VERSION=1.*
     - php: 7.3
+      env: MONOLOG_VERSION=1.*
+    - php: 7.4
+      env: MONOLOG_VERSION=1.*
+    # Test against seldaek/monolog 2.x
+    - php: 7.2
+      env: MONOLOG_VERSION=2.*
+    - php: 7.3
+      env: MONOLOG_VERSION=2.*
+    - php: 7.4
+      env: MONOLOG_VERSION=2.*
     # Test against dev versions
     - php: nightly
-      env: DEPENDENCIES=dev
+      env: DEPENDENCIES=dev MONOLOG_VERSION=1.*
+    - php: nightly
+      env: DEPENDENCIES=dev MONOLOG_VERSION=2.*
   allow_failures:
-    - env: DEPENDENCIES=dev
+    - php: nightly
 
 before_install:
     - composer self-update
     - if [ "$DEPENDENCIES" = "dev" ]; then perl -pi -e 's/^}$/,"minimum-stability":"dev"}/' composer.json; fi;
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -24,8 +24,20 @@ matrix:
     - php: 7.1
       # There is a bug in PHPUnit 5.7
       env: SYMFONY_PHPUNIT_VERSION=6.5
+    # Test against seldaek/monolog 1.x
```
stof, line 27: `--prefer-lowest` is not about latest monolog 1

Tests/DependencyInjection/MonologExtensionTest.php
```diff
@@ -292,6 +295,9 @@ public function testRavenHandlerWhenConfigurationIsWrong()
         }
     }
 
+    /**
+     * @group version1only
```
stof, line 299: Wouldn't it be better to detect the monolog version in the test and skip them instead of relying on the caller to exclude a group ?

Tests/ForwardCompatTestTrait.php
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MonologBundle\Tests;
+
+use PHPUnit\Framework\TestCase;
+
+// Auto-adapt to PHPUnit 8 that added a `void` return-type to the setUp/tearDown methods
```
stof, line 16: I suggest using the phpunti-bridge instead, which deals with this

Tests/DependencyInjection/MonologExtensionTest.php
dbrumann: Yes, I am currently working on replacing this with a `markTestSkipped` where I look for the handler class from Monolog 1 instead. This way the test will also be more reliable when running locally without the flag

Tests/ForwardCompatTestTrait.php
dbrumann: I couldn't find the trait in the bridge, but I will look again.

dbrumann: I have restored the custom Trait as tests with --prefer-lowest will fetch a version of the bridge, which does not have the trait.

xabbuh (approved):

Tests/DependencyInjection/Compiler/AddSwiftMailerTransportPassTest.php
```diff
@@ -20,13 +21,15 @@
  */
 class AddSwiftMailerTransportPassTest extends TestCase
 {
+    use SetUpTearDownTrait;
```
stof, line 24: a native solution could be to use methods tagged as `@before` instead, which works since a very long time in PHPUnit.

fabpot (approved):

lyrixx (approved):

lyrixx:
Thank you @dbrumann.

PR #370, opened 2020-10-27
Drop support Symfony 3.4

### LINKED ISSUE #365: 3.6.0 causes issues with Symfony 3.4 and custom log handler services
Since 3.6.0 we're encountering the following issue, but only when using with Symfony v3.4.45, with v4.4.15 all is working fine: 

```text
Fatal error: Uncaught Symfony\Component\DependencyInjection\Exception\InvalidArgumentException: Unused binding "Symfony\Contracts\HttpClient\HttpClientInterface" in service "instanceof.Monolog\Handler\HandlerInterface.0.Pimcore\Log\Handler\ApplicationLoggerDb". in /home/travis/build/pimcore/pimcore/vendor/symfony/symfony/src/Symfony/Component/DependencyInjection/Compiler/ResolveBindingsPass.php on line 50
Symfony\Component\DependencyInjection\Exception\InvalidArgumentException: Unused binding "Symfony\Contracts\HttpClient\HttpClientInterface" in service "instanceof.Monolog\Handler\HandlerInterface.0.Pimcore\Log\Handler\ApplicationLoggerDb". in /home/travis/build/pimcore/pimcore/vendor/symfony/symfony/src/Symfony/Component/DependencyInjection/Compiler/ResolveBindingsPass.php on line 50
```

This is the used service definition: 
https://github.com/pimcore/pimcore/blob/416ed369e071cfab175b33b9a4596c130dbe1a29/bundles/CoreBundle/Resources/config/logging.yml#L33

This all seems to be related with this lines: 
https://github.com/symfony/monolog-bundle/blob/master/DependencyInjection/MonologExtension.php#L158 

But I've currently no idea why this breaks the service definition in Symfony 3.4

lyrixx:
fixes #365

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,15 +15,10 @@ env:
 matrix:
   fast_finish: true
   include:
-    - php: 5.6
-      env: COMPOSER_FLAGS="--prefer-lowest" SYMFONY_DEPRECATIONS_HELPER=weak
-    # Test against Symfony LTS versions
-    - php: 5.6
-      env: SYMFONY_VERSION="3.4.*"
-    - php: 7.0
     - php: 7.1
-      # There is a bug in PHPUnit 5.7
-      env: SYMFONY_PHPUNIT_VERSION=6.5
+      env: COMPOSER_FLAGS="--prefer-lowest" SYMFONY_DEPRECATIONS_HELPER=weak
+    - php: 7.1
+      env: SYMFONY_VERSION="4.4.*"
     # Test against seldaek/monolog 1.x
     - php: 7.2
       env: MONOLOG_VERSION=1.*
--- a/composer.json
+++ b/composer.json
@@ -3,7 +3,7 @@
     "type": "symfony-bundle",
     "description": "Symfony MonologBundle",
     "keywords": ["log", "logging"],
-    "homepage": "http://symfony.com",
+    "homepage": "https://symfony.com",
     "license": "MIT",
     "authors": [
         {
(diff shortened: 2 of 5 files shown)


derrabus (approved):

dbrumann (approved):

.travis.yml
```diff
@@ -15,12 +15,8 @@ env:
 matrix:
   fast_finish: true
   include:
-    - php: 5.6
-      env: COMPOSER_FLAGS="--prefer-lowest" SYMFONY_DEPRECATIONS_HELPER=weak
```
jderusse, line 19: Shouldn't `--prefer-lowest` tests be moved to another test suite?

derrabus: Oh right, good catch.

.travis.yml
```diff
@@ -15,12 +15,8 @@ env:
...
-    # Test against Symfony LTS versions
-    - php: 5.6
-      env: SYMFONY_VERSION="3.4.*"
     - php: 7.0
+      env: SYMFONY_VERSION="4.4.*"
```
derrabus, line 19: Symfony 4.4 is not compatible with php 7.0.

.travis.yml
```diff
@@ -15,12 +15,8 @@ env:
...
     - php: 7.1
       # There is a bug in PHPUnit 5.7
       env: SYMFONY_PHPUNIT_VERSION=6.5
```
derrabus, line 22: Can we bump to `7.5` here?

composer.json
```diff
@@ -12,20 +12,20 @@
...
             "name": "Symfony Community",
-            "homepage": "http://symfony.com/contributors"
+            "homepage": "https://symfony.com/contributors"
         }
     ],
     "require": {
-        "php": ">=5.6",
-        "symfony/monolog-bridge": "~3.4 || ~4.0 || ^5.0",
+        "php": ">=7.1.3",
+        "symfony/monolog-bridge": "~4.0 || ^5.0",
```
derrabus, line 20: I think we can be bold here.

```suggestion
        "symfony/monolog-bridge": "~4.4 || ^5.0",
```

The feature set of 3.4 and 4.0 is equivalent, so we won't gain much from dropping 3.4 alone.

derrabus (changes requested):

composer.json
stof: agreed. And 3.4 still receive security fixes while 4.3 already stopped a while ago, so there is no reason to keep support 4.3 once we drop 3.4

composer.json
```diff
@@ -12,20 +12,20 @@
...
+        "symfony/config": "~4.0 || ^5.0",
+        "symfony/http-kernel": "~4.0 || ^5.0",
         "monolog/monolog": "~1.22 || ~2.0"
     },
     "require-dev": {
-        "symfony/yaml": "~3.4 || ~4.0 || ^5.0",
-        "symfony/console": "~3.4 || ~4.0 || ^5.0",
+        "symfony/yaml": "~4.0 || ^5.0",
+        "symfony/console": "~4.0 || ^5.0",
         "symfony/phpunit-bridge": "^4.4 || ^5.0"
```
stof, line 29: `symfony/phpunit-bridge` should be changed to `^5.1`. there is no point in using an older version there, given that 5.1 supports all PHP versions we need

.travis.yml
lyrixx: We should make a choice, because here we have two ways to do the same thing
--prefer-lower (so SF 4.4) or SYMFONY_VERSION=4.4

lyrixx: I choose prefer-lowest

derrabus: Those are actually two different things. With `--prefer-lowest` we would test against Symfony 4.4.0, but there should also be a test run with the latest 4.4 release. I think we need both.

lyrixx: ok, I'll do that 👍🏼

lyrixx:
It should be OK now 💚

derrabus (approved):

jderusse (approved):
I'm :+1: for this PR from a technical point of view, 

But I wonder what is the benefit for this package to drop support for the outdated version of Symfony?
Do we expect to use php 7.1 features here?

lyrixx:
IIRC, I did this pr to fix https://github.com/symfony/monolog-bundle/issues/365#issuecomment-705417779

derrabus:
At the moment, php 5 support is dropped from other repos as well (`symfony/symfony`, `symfony/polyfill`) and I don't think we're doing ourselves a favor by maintaining php 5 support for future releases. That being said, old php 5 compatible releases of this package won't disappear after merging this PR.

php 5 is EOL for almost two years now. Let's move on.

> Do we expect to use php 7.1 features here?

We should start to. 😃

Issue #367, opened 2020-10-10
Segmentation fault (core dumped) with symfony_mailer handler


ahmed-bhs:
Q | A
-- | --
Bug report? | yes
Feature request? | no
BC Break report? | no
RFC? | no
Symfony version | 4.4.13
monolog/monolog | 2.1.1
symfony/monolog-bridge | v5.1.1
symfony/monolog-bundle | v3.6.0

I'm trying  to configure Monolog to send an email when an error occurs within my application. This configuration looks correct but every time I launch bin/console cache:clear, I get the "Segmentation fault (core dumped)" error.

This is my monolog config:

        main:
            type:         fingers_crossed
            action_level: critical
            handler:      deduplicated
            channels: ["!event", "!console"]
        deduplicated:
            type:    deduplication
            time: 10
            handler: symfony_mailer
        symfony_mailer:
            type:       symfony_mailer
            from_email: '%env(SENDER)%'
            to_email:   'my mail@gmail.com'
            subject:    'An Error Occurred! '
            level:      critical
            formatter:  monolog.formatter.html
            content_type: text/html

Any help would be greatly appreciated!
If you need more information, just let me know :)

lyrixx:
Hello, could you create a small reproducer? Thanks

adrien-f:
👋 Hey I think we're hitting that issue as well, I've dumped the core and looked into it with GDB: 

```
gdb) zbacktrace
[0x7faeaf722560] preg_replace_callback("/%%|%([^%\s]+)%/", object[0x7faeaf7225c0], "warn") [internal function]
[0x7faeaf7224b0] Symfony\Component\DependencyInjection\ParameterBag\ParameterBag->resolveString("warn", array(0)[0x7faeaf722510]) /var/www/vendor/symfony/dependency-injection/ParameterBag/ParameterBag.php:219
[0x7faeaf7223e0] Symfony\Component\DependencyInjection\ParameterBag\ParameterBag->resolveValue("warn") /var/www/vendor/symfony/dependency-injection/ParameterBag/ParameterBag.php:187
[0x7faeaf722280] Symfony\Component\DependencyInjection\ContainerBuilder->resolveEnvPlaceholders("warn", true, reference) /var/www/vendor/symfony/dependency-injection/ContainerBuilder.php:1416
[...]

PR #373, opened 2020-11-09
Fix infinite loop in LogLevel detection

jderusse:
While converting the logLevel into the right const value, calling `$container->resolveEnvPlaceholders($level, true);` does not always throw an exception and returns the value provided in INPUT, leading to an infinite loop

This PR fixes  https://github.com/symfony/monolog-bundle/issues/367#issuecomment-723988352 BUT is not related to the issue https://github.com/symfony/monolog-bundle/issues/367#issue-718712384

### DIFF
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -62,7 +62,7 @@ class MonologExtension extends Extension
             throw new \InvalidArgumentException(sprintf('Could not match "%s" to a log level.', $level));
         }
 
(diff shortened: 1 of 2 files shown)


derrabus (approved):

greg0ire (approved):

michelv (approved):

lyrixx (approved):

lyrixx:
You were fast on this one, thanks Jérémy.

PR #377, opened 2020-11-24
Ensure all handlers are explicitly closed on kernel shutdown

andy-educake:


MonologBundle.php
```diff
@@ -52,4 +52,15 @@ public static function includeStacktraces(HandlerInterface $handler)
...
+    public function shutdown()
+    {
+        parent::shutdown();
+
+        $handlerManager = $this->container->get('monolog.handler_manager');
+
+        $handlerManager->close();
+    }
+
+
```
lyrixx, line 65: Extra \n to remove

Services/HandlerManager.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+
+namespace Symfony\Bundle\MonologBundle\Services;
+
+
+use Monolog\Handler\HandlerInterface;
+
+class HandlerManager
```
lyrixx, line 9: I don't like `*Manager` class name. Could you rename it to `HandlerHolder`, or something similar ?

Services/HandlerManager.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+
+namespace Symfony\Bundle\MonologBundle\Services;
```
lyrixx, line 4: We usually don't use `Services` namespace. Can you put this class in `Symfony\Bundle\MonologBundle` instead?

Services/HandlerManager.php
```diff
@@ -0,0 +1,28 @@
...
+
+
+use Monolog\Handler\HandlerInterface;
+
+class HandlerManager
+{
+
```
lyrixx, line 11: Extra new lines

Resources/config/monolog.xml
```diff
@@ -40,5 +40,9 @@
         <service id="monolog.http_client" class="Symfony\Contracts\HttpClient\HttpClientInterface" public="false">
             <factory class="Symfony\Component\HttpClient\HttpClient" method="create" />
         </service>
+
+        <service id="monolog.handler_manager" class="Symfony\Bundle\MonologBundle\Services\HandlerManager" public="false">
+            <argument type="tagged" tag="monolog.handler" />
```
lyrixx, line 45: tagged_iterator?

Services/HandlerManager.php
```diff
@@ -0,0 +1,28 @@
...
+    /**
+     * @var HandlerInterface[]
+     */
+    private $handlers;
+
+    public function __construct(array $handlers)
```
lyrixx, line 17: I don't think this is an array. I think it should be iterable instead

lyrixx (commented):
I like this PR, but I'm not sure about it. If we do that, we should do the same thing for all others bundle that does IO.

@nicolas-grekas WDYT?

Services/HandlerManager.php
andy-educake: Yes, agree that `Manager` is not very decriptive, not sure `Holder` is any more descriptive though. Have changed it to `HandlerLifecycleManager` as that feels closer to what it actually does. Is that OK, or do you have other ideas?

andy-educake:
Thanks for the comments @lyrixx Have dealt with all of them I think. I notice that Travis has failed for PHP 5.6 and PHP 7.0. Possibly because I used the `iterable` type hint on the constructor param. Should I use the `Iterator` interface instead for bc?

Services/HandlerManager.php
lyrixx: Very good name 👍🏼

HandlerLifecycleManager.php
```diff
@@ -0,0 +1,27 @@
...
+use Monolog\Handler\HandlerInterface;
+
+class HandlerLifecycleManager
+{
+    /**
+     * @var HandlerInterface[]
+     */
+    private $handlers;
+
+    public function __construct(\IteratorAggregate $handlers)
```
derrabus, line 16:
```suggestion
    public function __construct(iterable $handlers)
```

The type `IteratorAggregate` is a bit more specific that it needs to be, isn‘t it?

Resources/config/monolog.xml
```diff
@@ -40,5 +40,9 @@
         <service id="monolog.http_client" class="Symfony\Contracts\HttpClient\HttpClientInterface" public="false">
             <factory class="Symfony\Component\HttpClient\HttpClient" method="create" />
         </service>
+
+        <service id="monolog.handler_manager" class="Symfony\Bundle\MonologBundle\HandlerLifecycleManager" public="true">
+            <argument type="tagged_iterator" tag="monolog.handler" />
```
stof, line 45: Rather than using `type="tagged_iterator"`, you should use a compiler pass to create the IteratorArgument (or drop the tag entirely and manage everything in the DI extension as I don't think we need to support handlers added elsewhere). This way, you can use weak references in the IteratorArgument, which will skip services which haven't been instantiated yet

HandlerLifecycleManager.php
stof: the bundle supports PHP 5.6+ currently

stof:
Regarding a generic solution in symfony, I think we need something similar to the `ServiceResetter`, but which is always called by the kernel on `shutdown`, for services which need to close some resources for their shutdown. This way, MonologBundle could hook the `close` method of handlers in there, DoctrineBundle could hook connections, etc...

HandlerLifecycleManager.php
stof: but `Traversable` should be used instead of `IteratorAggregate`

HandlerLifecycleManager.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+
+namespace Symfony\Bundle\MonologBundle;
+
+
+use Monolog\Handler\HandlerInterface;
+
+class HandlerLifecycleManager
```
stof, line 9: should be tagged as `@internal`. We really don't want to expose it in the BC promise.

HandlerLifecycleManager.php
lyrixx: we should drop support for SF 3.X, and bump PHP to 7.1
IMHO, we must do that before this PR

Resources/config/monolog.xml
andy-educake: I'm not sure I'm familiar enough with the Container to change this right now  but given https://github.com/symfony/monolog-bundle/pull/377#discussion_r531587372 I should probably pause on this PR now anyway?

HandlerLifecycleManager.php
```diff
@@ -0,0 +1,27 @@
...
+
+    public function __construct(\IteratorAggregate $handlers)
+    {
+        $this->handlers = $handlers;
+    }
+
+    public function close()
+    {
+        foreach ($this->handlers as $handler) {
+            $handler->close();
```
GromNaN, line 24: This will instantiate all the handlers, even those that have not been used.

MonologBundle.php
```diff
@@ -52,4 +52,14 @@ public static function includeStacktraces(HandlerInterface $handler)
...
         }
     }
+
+    public function shutdown()
+    {
+        parent::shutdown();
+
+        $handlerManager = $this->container->get('monolog.handler_manager');
+
+        $handlerManager->close();
```
GromNaN, line 62:
```suggestion
        $this->container->get('monolog.handler_manager')->close();
```

HandlerLifecycleManager.php
GromNaN:
```suggestion
/**
 * @internal
 */
class HandlerLifecycleManager
```

GromNaN (commented):
Is it still necessary?

GromNaN:
Rebased the branch onto current 4.x and applied the standing review feedback so it can move forward. The previous structure (root `DependencyInjection/`, `Resources/config/monolog.xml`, root `MonologBundle.php`) no longer exists on 4.x, so this is a re-apply rather than a mechanical rebase.

Changes:
- `HandlerLifecycleManager` is now `@internal` (stof, GromNaN) and lives in `src/`.
- The constructor takes `iterable $handlers` (derrabus).
- `MonologBundle::shutdown()` uses the one-liner form (GromNaN), with a guard for when the manager is not defined.
- All handlers (including nested ones) are tagged `monolog.handler`. Nested handlers are excluded from `kernel.reset`, but they hold real resources and must be closed on shutdown.

stof suggestion implemented: instead of a `tagged_iterator` (which would instantiate every handler on shutdown, opening connections/files for unused handlers just to close them), a new `AddHandlersToManagerPass` compiler pass injects the handlers as an `IteratorArgument` of weak references (`ContainerInterface::IGNORE_ON_UNINITIALIZED_REFERENCE`). This is the same mechanism Symfony uses for `services_resetter`: the generated iterator only yields already-instantiated services, so unused handlers are never instantiated at shutdown.

Added a CHANGELOG entry and tests (handlers are tagged; the manager receives weak references).

Local checks: full PHPUnit suite passes, php-cs-fixer is clean.

Happy to adjust anything.

GromNaN:
Regarding stof's earlier point about a generic Symfony solution: a framework-level mechanism similar to `services_resetter`, but invoked on kernel shutdown (so MonologBundle could hook handler `close()`, DoctrineBundle could hook connections, etc.), would be the ideal long-term answer. Symfony does not provide such a mechanism today: there is no `kernel.shutdown` event and `Kernel::shutdown()` only calls `Bundle::shutdown()`. So this PR uses `Bundle::shutdown()` as the available hook.

The design is intended to be forward-compatible with that future generic closer: handlers are tagged (`monolog.handler`) and collected by a compiler pass into an iterator of weak references. Swapping the bundle-specific `monolog.handler_lifecycle_manager` service for a framework-provided closer later would not require changing how handlers are tagged or collected.

Side note on the service being public: it has to be public because `Bundle::shutdown()` reaches it through `$this->container->get()`, the same way Symfony's `services_resetter` is public and fetched in `Kernel::boot()`. Private services are not retrievable via `get()` at runtime, and there is no shutdown event to subscribe to instead.

GromNaN:
Thanks @andy-educake for working on this feature, this is much appreciated.

PR #384, opened 2021-01-20
Typo in console formatter option name

HypeMC:
The `console_formater_options` has a typo. This PR deprecates the option and replaces it with `console_formatter_options` while maintaining backwards compatibility.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -38,7 +38,7 @@ use Monolog\Logger;
  *   - [verbosity_levels]: level => verbosity configuration
  *   - [level]: level name or int value, defaults to DEBUG
  *   - [bubble]: bool, defaults to true
- *   - [console_formater_options]: array
+ *   - [console_formatter_options]: array
  *
  * - firephp:
  *   - [level]: level name or int value, defaults to DEBUG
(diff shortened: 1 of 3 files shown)


dbrumann (approved):

OskarStark (approved):

jderusse (approved):

DependencyInjection/Configuration.php
```diff
@@ -946,6 +954,17 @@ public function getConfigTreeBuilder()
                             ->ifTrue(function ($v) { return 'predis' === $v['type'] && empty($v['redis']); })
                             ->thenInvalid('The host has to be specified to use a RedisLogHandler')
                         ->end()
+                        ->validate()
```
stof, line 957: this should be done in `beforeNormalization`, not in `validate` IMO, so that the copy of the options to the new name is done before merging various configs together. This ensures that each config file can migrate independently, instead of loosing all non-migrated config as soon as one file gets migrated.

Tests/DependencyInjection/ConfigurationTest.php
```diff
@@ -406,6 +407,46 @@ public function testWithRedisHandler()
         $this->assertEquals('monolog_redis_test', $config['handlers']['redis']['redis']['key_name']);
     }
 
+    /**
+     * @group legacy
+     */
+    public function testConsoleFormatterOptionsRename()
```
stof, line 413: please add another legacy test involving 2 configs, one migrated and one not migrated for the same handler.

stof: or maybe just another handler in this test

stof (changes requested):

DependencyInjection/Configuration.php
HypeMC: Makes sense, switched to `beforeNormalization`.

Tests/DependencyInjection/ConfigurationTest.php
HypeMC: Test updated

DependencyInjection/Configuration.php
```diff
@@ -783,6 +790,17 @@ public function getConfigTreeBuilder()
                             ->scalarNode('formatter')->end()
                             ->booleanNode('nested')->defaultFalse()->end()
                         ->end()
+                        ->beforeNormalization()
+                            ->always(static function ($v) {
+                                if (empty($v['console_formatter_options']) && !empty($v['console_formater_options'])) {
+                                    $v['console_formatter_options'] = $v['console_formater_options'];
+                                }
+
+                                unset($v['console_formater_options']);
```
stof, line 799: if you unset the node here, the deprecation won't ever be triggered by the node itself

HypeMC: Fixed

jderusse (approved):

maxhelias (approved):

jderusse:
Thank you @HypeMC.

PR #336, opened 2020-02-16
Handle composed monolog-bridge activation strategy

pvgnd:
Related to https://github.com/symfony/symfony/pull/35740

This PR instantiate the new HttpCodeActivationStrategy & NotFoundActivationStrategy with an inner strategy instead of an actionLevel

closes #369

PR #385, opened 2021-01-23
Fix deprecated

### LINKED ISSUE #369: Handle deprecation of the `actionLevel` argument in finger crossed strategies
Form Symfony 6, the `actionLevel` must be an inner strategy instead of plain old value / string / int.

To prevent the deprecation, when creating the definition of the bridge's strategy, we should probably wrap the action level in an inner strategy as is done in the bridges ?

See https://github.com/symfony/monolog-bridge/blob/5.x/Handler/FingersCrossed/HttpCodeActivationStrategy.php#L39-L44 and https://github.com/symfony/monolog-bridge/blob/5.x/Handler/FingersCrossed/NotFoundActivationStrategy.php#L39-L44

pvgnd:
closes #369

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Register resettable processors (`ResettableInterface`) for autoconfiguration (tag: `kernel.reset`)
 * Drop support for Symfony 3.4
+* Handle composed monolog-bridge activation strategy
 
 ## 3.6.0 (2020-10-06)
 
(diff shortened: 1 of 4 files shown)


jderusse:
Is it a duplicate of #336?

pvgnd:
> Is it a duplicate of #336?

Yes, as asked by @OskarStark https://github.com/symfony/monolog-bundle/pull/336#discussion_r561240019

jderusse:
I think there is a misunderstood. You don't need to open a new PR to trigger CI. You can push a new commit for this.

Anyway, I'm not sure this repo has fabbot configured (looking at other PR none, have such build). I'm closing this PR in favor of #336

Issue #335, opened 2019-12-05
Updating from symfony 4 to 5 breaks with null handler


gmponos:
I have upgraded my application from 4.4.1 to 5.0.1

inside my packages/test/monolog.yaml I have:
```
monolog:
    handlers:
        main:
            type: 'null'
            level: debug
```

When I run my tests I get `Error : Call to undefined method Monolog\Handler\NullHandler::pushProcessor()`

is this a bug or there is a workaround (besides setting test/debug handler)?

acasademont:
I downgraded to monolog 1.X in the meantime, that should fix it.

By default, if you only have `monolog-bridge` in your composer.json, it will pick up the highest monolog version available (2.0.5). Explicitly requiring `monolog/monolog:^1.25` should do it

tzunghaor:
Symfony tried to push psr formatter processor to my null handler, but explicitly disabling it solved my problem:

```yaml
monolog:
    handlers:
        main:
            type: 'null'
            process_psr_3_messages: false
```

Seldaek:
The bundle should check whether ProcessableHandlerInterface is implemented (in Monolog 2) before calling pushProcessor.

jderusse:
Fixed by #390

PR #386, opened 2021-01-30
Issue 335

VBelozyorov:
Since Monolog v2 not all handlers has `pushProcessor()` method, for example `Monolog\Handler\NullHandler`. Thus seems we should check it before add method call to handler's service definition.

Ref #335

### DIFF
--- a/DependencyInjection/Compiler/AddProcessorsPass.php
+++ b/DependencyInjection/Compiler/AddProcessorsPass.php
@@ -11,8 +11,9 @@
 
 namespace Symfony\Bundle\MonologBundle\DependencyInjection\Compiler;
 
-use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Monolog\Handler\ProcessableHandlerInterface;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Reference;
 
 /**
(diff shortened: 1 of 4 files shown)


VBelozyorov:
I've faced with some troubles. 

Due `Monolog\Handler\ProcessableHandlerInterface` was backported to v1 (https://github.com/Seldaek/monolog/commit/6d76eaaf7699d6bab1609b0a5abcc9d2326ad566) I'm cannot use existance of this interface as logical switch between Monolog v1 and v2. 

I've tried to use Reflection for this switch, but for `NullHandler` got `%monolog.handler.null.class%` from `$definition->getClass()` and stuck for now.

@lyrixx @Seldaek maybe you guys can give me advice? Thanks in advance!

Seldaek:
You can use this to detect monolog version https://github.com/Seldaek/monolog/blob/main/src/Monolog/Logger.php#L88

DependencyInjection/Compiler/AddProcessorsPass.php
```diff
@@ -46,13 +48,15 @@ public function process(ContainerBuilder $container)
...
-                    $processor = new Reference($id);
+                if (1 === Logger::API || is_a($definition->getClass(), ProcessableHandlerInterface::class, true)) {
+                    if (!empty($tag['method'])) {
+                        $processor = [new Reference($id), $tag['method']];
+                    } else {
+                        // If no method is defined, fallback to use __invoke
+                        $processor = new Reference($id);
+                    }
+                    $definition->addMethodCall('pushProcessor', [$processor]);
                 }
```
VBelozyorov, line 59: With Monolog v2, should we throw Exception if handler doesn't implements `ProcessableHandlerInterface` here?

DependencyInjection/Compiler/AddProcessorsPass.php
```diff
@@ -46,13 +48,15 @@ public function process(ContainerBuilder $container)
                     $definition = $container->getDefinition('monolog.logger_prototype');
                 }
 
-                if (!empty($tag['method'])) {
-                    $processor = [new Reference($id), $tag['method']];
-                } else {
-                    // If no method is defined, fallback to use __invoke
-                    $processor = new Reference($id);
+                if (1 === Logger::API || is_a($definition->getClass(), ProcessableHandlerInterface::class, true)) {
```
jderusse, line 51: in case `empty($tag['handler'])`, `$definition` might contain an instance of `Logger` which is neither Version1 nor `ProcessableHandlerInterface`.

jderusse, line 51: `$definition->getClass()` may be null (childDefinition), you have to check the parent class.

jderusse (commented):
Sorry, I missed this PR when I created https://github.com/symfony/monolog-bundle/pull/390

jderusse:
Fixed by #390

PR #388, opened 2021-02-14
Use `ActivationStrategy` instead of `actionLevel` when available

### LINKED ISSUE #369: Handle deprecation of the `actionLevel` argument in finger crossed strategies

jderusse:
Take over #336

Fixes #369

### DIFF
--- a/Tests/DependencyInjection/FixtureMonologExtensionTest.php
+++ b/Tests/DependencyInjection/FixtureMonologExtensionTest.php
@@ -11,15 +11,36 @@
 
 namespace Symfony\Bundle\MonologBundle\Tests\DependencyInjection;
 
-use Symfony\Bundle\MonologBundle\DependencyInjection\MonologExtension;
+use Monolog\Handler\FingersCrossed\ErrorLevelActivationStrategy;
+use Symfony\Bridge\Monolog\Processor\SwitchUserTokenProcessor;
 use Symfony\Bundle\MonologBundle\DependencyInjection\Compiler\LoggerChannelPass;
+use Symfony\Bundle\MonologBundle\DependencyInjection\MonologExtension;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Reference;
 
 abstract class FixtureMonologExtensionTest extends DependencyInjectionTest
 {
+    /** @group legacy */
+    public function testLegacyLoadWithSeveralHandlers()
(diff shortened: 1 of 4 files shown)


Tests/DependencyInjection/FixtureMonologExtensionTest.php
```diff
@@ -11,15 +11,36 @@
...
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Reference;
 
 abstract class FixtureMonologExtensionTest extends DependencyInjectionTest
 {
+    /** @group legacy */
+    public function testLegacyLoadWithSeveralHandlers()
+    {
+        if (class_exists(SwitchUserTokenProcessor::class)) {
+            $this->markTestSkipped('Symfony Monolog < 5.2 is needed.');
```
derrabus, line 28:
```suggestion
            $this->markTestSkipped('Symfony MonologBridge < 5.2 is needed.');
```

derrabus (approved):

Tests/DependencyInjection/FixtureMonologExtensionTest.php
derrabus: There are more places like this one. 😃

maxhelias (approved):

Jean85 (approved):

stof (approved):

stephanvierkant:
Thanks! Any idea when a new tagged version will be released?

jokaorgua:
@jderusse @stof would be really nice if you could give us ETA for a new tag with these fixes

jokaorgua:
@jderusse @stof any ETA?

patrykmartynowicz:
@jokaorgua As I see it's available already at `v3.7.0`.

jokaorgua:
@patrykmartynowicz thanks. but due to some reason, GitHub marks 3.6.0 as the latest release

derrabus:
> GitHub marks 3.6.0 as the latest release

Fixed. 😃

PR #390, opened 2021-02-15
Fix call to handler::pushProcessor

jderusse:
This PR fixes call to undefined method `pushProcessor` on handler that does not implements the `ProcessableHandlerInterface`.

### DIFF
--- a/DependencyInjection/Compiler/AddProcessorsPass.php
+++ b/DependencyInjection/Compiler/AddProcessorsPass.php
@@ -11,8 +11,9 @@
 
 namespace Symfony\Bundle\MonologBundle\DependencyInjection\Compiler;
 
-use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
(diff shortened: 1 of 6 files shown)


stof (approved):

DependencyInjection/Compiler/AddProcessorsPass.php
```diff
@@ -36,6 +37,14 @@ public function process(ContainerBuilder $container)
 
                 if (!empty($tag['handler'])) {
                     $definition = $container->findDefinition(sprintf('monolog.handler.%s', $tag['handler']));
+                    $parentDef = $definition;
+                    while (!$parentDef->getClass() && $parentDef instanceof ChildDefinition) {
+                        $parentDef = $container->findDefinition($parentDef->getParent());
+                    }
+                    $class = $container->getParameterBag()->resolveValue($parentDef->getClass());
+                    if (!method_exists($class, 'pushProcessor')) {
+                        continue;
```
stof, line 46: should this case silently ignore the push or fail ? It might be unexpected to explicitly reference an handler and then have the config ignored.

jderusse: Yeah, indeed. Fixed.

PR #393, opened 2021-03-09
Add ability to specify custom Sentry hub service

HypeMC:
As mentioned in https://github.com/getsentry/sentry-docs/pull/3137#discussion_r588942726 , this will allow users to set a custom hub service, eg the one registered by the [Sentry bundle](https://github.com/getsentry/sentry-symfony).

PR #394, opened 2021-03-09
Add Sentry Hub instance in DI container to resolve event context issue

dedpikhto:
Fixing Symfony issue with different Hub's while decorating default Sentry handler for extended event context.

Common way to extend log event context is to decorate Sentry\Monolog\Handler:
```php
final class SentryContextHandlerDecorator extends AbstractProcessingHandler
{
    private Handler $handler;

    public function __construct(Handler $handler)
    {
        $this->handler = $handler;

        parent::__construct($handler->getLevel(), $handler->getBubble());
    }

    public function handle(array $record): bool
    {
        $result = false;

        withScope(function (Scope $scope) use ($record, &$result): void {
            if (isset($record['context']) && is_array($record['context']) && count($record['context']) > 0) {
                $scope->setContext('context', $record['context']);
            }

            $result = $this->handler->handle($record);
        });

        return $result;
    }
}
```

But if you do so with Symfony DI, you will get two instances of Hub: one from DI container, and one with SentrySdk in withScope(...) call. Changes to one Hub's scope won't affect second's hub scope.

This PR adds ability to use shared Hub in custom decorators:
```php
App\Logging\Handler\SentryContextHandlerDecorator:
    decorates: monolog.handler.sentry
    arguments:
        - '@App\Logging\Handler\SentryContextHandlerDecorator.inner'
        - '@monolog.handler.sentry.hub'
```
where `sentry` in `monolog.handler.sentry.hub` is name of handler.

Decorator with shared Hub instance:
```php
final class SentryContextHandlerDecorator extends AbstractProcessingHandler
{
    private Handler $handler;
    private Hub $hub;

    public function __construct(Handler $handler, Hub $hub)
    {
        $this->handler = $handler;
        $this->hub = $hub;

        parent::__construct($handler->getLevel(), $handler->getBubble());
    }

    public function handle(array $record): bool
    {
        $result = false;

        $this->hub->withScope(function (Scope $scope) use ($record, &$result): void {
            if (isset($record['context']) && is_array($record['context']) && count($record['context']) > 0) {
                $scope->setContext('context', $record['context']);
            }

            $result = $this->handler->handle($record);
        });

        return $result;
    }
}
```

More comments about problem: https://github.com/getsentry/sentry-php/pull/848#issuecomment-791346614

### DIFF
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -757,6 +757,7 @@ class MonologExtension extends Extension
                     'Sentry\\State\\Hub',
                     [new Reference($clientId)]
                 );
+                $container->setDefinition(sprintf('monolog.handler.%s.hub', $name), $hub);
 
                 // can't set the hub to the current hub, getting into a recursion otherwise...
                 //$hub->addMethodCall('setCurrent', array($hub));
--- a/Tests/DependencyInjection/MonologExtensionTest.php
+++ b/Tests/DependencyInjection/MonologExtensionTest.php
@@ -393,6 +393,7 @@ class MonologExtensionTest extends DependencyInjectionTest
         ]]]]);
         $this->assertTrue($container->hasDefinition('monolog.logger'));
         $this->assertTrue($container->hasDefinition('monolog.handler.sentry'));
+        $this->assertTrue($container->hasDefinition('monolog.handler.sentry.hub'));
 
         $logger = $container->getDefinition('monolog.logger');
         $this->assertDICDefinitionMethodCallAt(0, $logger, 'useMicrosecondTimestamps', ['%monolog.use_microseconds%']);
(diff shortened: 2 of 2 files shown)


dedpikhto:
@stof @jderusse could you review, please?

jderusse (approved):

jderusse:
I think the PR #393 would cover the same issue, but is more flexible.

dedpikhto:
> I think the PR #393 would cover the same issue, but is more flexible.

It has the same flexibility (i thins it's rare case, when you need overriden hub, but you can do it) and it's simplier for common use case because you don't need to describe your own hub

jderusse (approved):

chalasr (approved):

jderusse:
@dedpikhto could you please rebase your PR?

jderusse:
Thank you @dedpikhto.

PR #396, opened 2021-03-17
Fix deprecation

jderusse:


composer.json
```diff
@@ -19,7 +19,7 @@
         "php": ">=7.1.3",
         "symfony/monolog-bridge": "~4.4 || ^5.0",
         "symfony/dependency-injection": "^4.4 || ^5.0",
-        "symfony/config": "~4.4 || ^5.0",
+        "symfony/config": "^5.1",
```
jderusse, line 22: Any idea how to keep compatibility with 4.4 while signature of `setDeprecated` method changed in 5.1?

stof: See https://github.com/doctrine/DoctrineBundle/blob/708ed6f99c3d0142e15a8b05022946367cfcc075/DependencyInjection/Configuration.php#L722-L733 for the trick used in DoctrineBundle

@nicolas-grekas the impact on the ecosystem was not properly considered IMO, when prepending new arguments in the new signature.

HypeMC: @jderusse Maybe the [baseline feature](https://symfony.com/doc/current/components/phpunit_bridge.html#baseline-deprecations) could be used here. It would require bumping phpunit bridge to 5.2, but i don't think that would cause any problems.

stof: @HypeMC this would cause vendor deprecations in all projects using the bundle, which will then complain to us. Ignoring the deprecation in our own CI is not the solution there

jderusse: Thanks for the hint @stof

derrabus (approved):

PR #400, opened 2021-03-29
add alias for null handler - fixes #133

### LINKED ISSUE #133: Cannot use the NullHandler with XML configuration
Hi,

I'm working on a Symfony 2 application, and I'd like to temporary disable logging.
I found that adding the `NullHandler` on top of the configured handlers shall do the job.
Here's my monolog XML configuration:

``` xml
<monolog:config>
    <monolog:handler name="null_handler" type="null" />
    <monolog:handler name="file_handler" type="stream" level="info" path="%kernel.logs_dir%/%kernel.environment%.log" />
</monolog:config>
```

However, when I access a web page, this error occured:

```
InvalidArgumentException in MonologExtension.php line 611:
Invalid handler type "" given for handler "null_handler"
```

It seems the `null` type in the XML parameter is treated as a blank string, and I guess there may be an issue in the `Configuration` class of the bundle.
What is your opinion?

Thanks for your help,
BR

gggeek:
Fixes #133

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Use `ActivationStrategy` instead of `actionLevel` when available 
 * Register resettable processors (`ResettableInterface`) for autoconfiguration (tag: `kernel.reset`)
 * Drop support for Symfony 3.4
+* Support 'null_handler' as alias for 'null', to ease problems with yaml and xml configuration
 
 ## 3.6.0 (2020-10-06)
 
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -188,7 +188,7 @@ class MonologExtension extends Extension
         }
 
         if (null === $handler['process_psr_3_messages']) {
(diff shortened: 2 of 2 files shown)


jderusse (commented):
Could you please add tests?

In YAML I think you can escape the `'null'` to fix solve this, but indeed this is an issue with XML configuration.
I wonder if it worth adding a new specific handler while people can also use the `noop` handle to achieve almost the same thing..

gggeek:
> In YAML I think you can escape the 'null' to fix solve this

True, but it still probably bites many developers the first time that they try to set it up. They might find it easier to use `null_handler` than `'null'`

> I wonder if it worth adding a new specific handler while people can also use the noop handle to achieve almost the same thing..

I am not sure I get what you mean here... It seems to me that noop handler and null handler behave differently, as the latter is a blackhole sink, while the former is not

I'll take a look at tests - but I might need some support with those

GromNaN:
It took time, but it's now fixed by https://github.com/symfony/monolog-bundle/pull/548. Thanks for exploring solution.

PR #403, opened 2021-04-10
Register channels to ResetLoggersWorkerSubscriber

l-vo:
Channels registration for symfony/symfony#40761

### DIFF
--- a/MonologBundle.php
+++ b/MonologBundle.php
@@ -14,13 +14,16 @@ namespace Symfony\Bundle\MonologBundle;
 use Monolog\Formatter\JsonFormatter;
 use Monolog\Formatter\LineFormatter;
 use Monolog\Handler\HandlerInterface;
+use Symfony\Bridge\Monolog\Messenger\ResetLoggersWorkerSubscriber;
 use Symfony\Bundle\MonologBundle\DependencyInjection\Compiler\AddSwiftMailerTransportPass;
+use Symfony\Bundle\MonologBundle\DependencyInjection\Compiler\ChannelsToMessengerPass;
 use Symfony\Component\HttpKernel\Bundle\Bundle;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Bundle\MonologBundle\DependencyInjection\Compiler\LoggerChannelPass;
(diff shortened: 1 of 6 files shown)


fabpot:
The related PR in Symfony has been merged for 5.3.

MonologBundle.php
```diff
@@ -38,6 +41,10 @@ public function build(ContainerBuilder $container)
         $container->addCompilerPass(new FixEmptyLoggerPass($channelPass));
         $container->addCompilerPass(new AddProcessorsPass());
         $container->addCompilerPass(new AddSwiftMailerTransportPass());
+
+        if (class_exists(MessageBusInterface::class) && class_exists(ResetLoggersWorkerSubscriber::class)) {
```
chalasr, line 45: Only ResetLoggersWorkerSubscribed is needed, right?

l-vo: No, both are needed to me. `ResetLoggersWorkerSubscriber` is from MonologBridge. If Monolog bridge is installed but not Messenger, we don't want `ResetLoggersWorkerSubscriber` to be activated.

chalasr: Of course, thanks.

chalasr (approved):

lyrixx (approved):

alxndrbauer:
Is there any particular reason why this is not merged yet?

lyrixx:
I don't thin this is good actually. The listener will always be registered. That means when Symfony handle an http request, if a message is sent in a sync transport, monolog will be flushed. So if something goes wrong after that, all log about routing, security etc will be lost 😐

I created another pr to fix this issue more "globally" https://github.com/symfony/symfony/pull/41163

lyrixx:
@l-vo Thanks for working on this topic.

As I have explain it in my previous comment, the alternative implementation provided by symfony/symfony#41163  (and subsequent PR) is better.

I'm also [deprecating](https://github.com/symfony/symfony/issues/43271) the integration in the bridge

I hope your understand the move here.

So, I'm sorry for closing your PR, but I think it's better like this.

Again, thanks for your first PR.

l-vo:
@lyrixx No problem, the reason is legit, thank you 🙂

PR #414, opened 2021-07-29
Support for PHP 8

mradhi:
There is an issue when we try to install the monolog bundle with composer using PHP 8.0:

`- monolog/monolog[2.0.0, ..., 2.0.2] require php ^7.2 -> your php version (8.0.8) does not satisfy that requirement.`

So to fix that, we need to upgrade the `monolog/monolog` package from "~2.0" to the version "~2.1", which already added a tentative support for PHP 8.

Check the [changelog file ](https://github.com/Seldaek/monolog/blob/2.1.0/CHANGELOG.md)for more details about the new version.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "symfony/dependency-injection": "^4.4 || ^5.0 || ^6.0",
         "symfony/config": "~4.4 || ^5.0 || ^6.0",
         "symfony/http-kernel": "~4.4 || ^5.0 || ^6.0",
-        "monolog/monolog": "~1.22 || ~2.0"
+        "monolog/monolog": "~1.22 || ~2.1"
     },
     "require-dev": {
         "symfony/yaml": "~4.4 || ^5.0 || ^6.0",


composer.json
```diff
@@ -21,7 +21,7 @@
         "symfony/dependency-injection": "^4.4 || ^5.0 || ^6.0",
         "symfony/config": "~4.4 || ^5.0 || ^6.0",
         "symfony/http-kernel": "~4.4 || ^5.0 || ^6.0",
-        "monolog/monolog": "~1.22 || ~2.0"
+        "monolog/monolog": "~1.22 || ~2.1"
```
jderusse, line 24: This package already allow `monolog/monolog 2.1, this change won't fix your issue.

> ~2.0 is equivalent to >=2.0 <3.0.0

From composer documentation : https://getcomposer.org/doc/articles/versions.md#tilde-version-range-

mradhi: That's true ! sorry about that, I think this requires more investigation as the issue is related to other packages, this is the output of the composer:

```
  Problem 1
    - monolog/monolog[1.22.0, ..., 1.26.1] require psr/log ~1.0 -> found psr/log[1.0.0, ..., 1.1.4] but the package is fixed to 2.0.0 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command.
    - monolog/monolog[2.0.0, ..., 2.0.2] require php ^7.2 -> your php version (8.0.8) does not satisfy that requirement.
    - monolog/monolog[2.1.0, ..., 2.3.2] require psr/log ^1.0.1 -> found psr/log[1.0.1, ..., 1.1.4] but the package is fixed to 2.0.0 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command.
    - symfony/monolog-bundle v3.7.0 requires monolog/monolog ~1.22 || ~2.0 -> satisfiable by monolog/monolog[1.22.0, ..., 1.26.1, 2.0.0, ..., 2.3.2].
    - Root composer.json requires symfony/monolog-bundle ^3.7 -> satisfiable by symfony/monolog-bundle[v3.7.0].
```

I think it's better to open a new issue regarding the support of PHP 8.0.

I'll close this PR.

derrabus: PHP 8 support is not your problem. You need to downgrade `psr/log` in order to install Monolog.

derrabus: See also Seldaek/monolog#1564

mradhi: @derrabus thank you, I downgraded the `psr/log` to 1.1.4 and it works !

PR #416, opened 2021-09-16
integrate Browser Console in Symfony

### LINKED ISSUE #152: BrowserConsoleHandler no output with php-fpm
The default BrowserConsoleHandler from monolog outputs by registering the send() method as a shutdown handler.

https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/BrowserConsoleHandler.php#L51

Unfortunately, when you use fpm, HttpFoundation/Response will issue a fastcgi_finish_request() (which immediately returns the content to e.g. nginx and closes the connection).
https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/Response.php#L373

So the javascript content is echoed after the content has been sent. Can be easily tested by disabling the  fastcgi_finish_request() line in HttpFoundation.

Should it be done maybe like the chromephp handler? Extending the monolog class, putting it into the bridge and overwrite the register_shutdown and send method with a kernel.response (onKernelResponse) event?

Unfortunately i have no time currently for a patch, we launch a new project in 2 weeks :)

Warxcell:
Fixes #152 

To be merged after https://github.com/symfony/symfony/pull/43065 is merged & released

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,8 @@
     "require-dev": {
         "symfony/yaml": "~4.4 || ^5.0 || ^6.0",
         "symfony/console": "~4.4 || ^5.0 || ^6.0",
-        "symfony/phpunit-bridge": "^5.2 || ^6.0"
+        "symfony/phpunit-bridge": "^5.2 || ^6.0",
+        "phpunit/phpunit": "^9.5"
     },
     "autoload": {
         "psr-4": { "Symfony\\Bundle\\MonologBundle\\": "" },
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -26,7 +26,8 @@
     "require-dev": {
         "symfony/yaml": "~4.4 || ^5.0 || ^6.0",
         "symfony/console": "~4.4 || ^5.0 || ^6.0",
-        "symfony/phpunit-bridge": "^5.2 || ^6.0"
+        "symfony/phpunit-bridge": "^5.2 || ^6.0",
+        "phpunit/phpunit": "^9.5"
```
HypeMC, line 30: This isn't needed since phpunit is downloaded via the `simple-phpunit` script provided by the `phpunit-bridge`.

Warxcell:
No longer needed.

PR #418, opened 2021-09-25
Switch CI to GitHub actions instead of Travis

HypeMC:
Switch CI to GitHub actions instead of Travis.


chalasr (approved):

derrabus (changes requested):
Let's have a green CI before we merge this.

https://github.com/HypeMC/monolog-bundle/actions/runs/1273089757

stof:
Running tests on PHP 8.1 requires using PHPUnit 9.5 rather than 9.4 to avoid the deprecations

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,62 @@
...
+                run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+            -   name: Cache dependencies
+                uses: actions/cache@v2
+                with:
+                    path: ${{ steps.composer-cache.outputs.dir }}
+                    key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.json') }}
+                    restore-keys: ${{ runner.os }}-composer-
+
+            -   name: Set composer minimum stability to dev
```
stof, line 47: Would be better described as `Configure composer minimum stability` as it supports setting it to something else than `dev`

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,62 @@
...
+on:
+    push:
+    pull_request:
+
+jobs:
+    test:
+        runs-on: ubuntu-latest
+        name: PHPUnit Tests
+        strategy:
+            fail-fast: true
```
stof, line 12: should be removed IMO. If tests fail on one of the jobs of the matrix, we still want to run the others to know whether the failure is only for that combination or no.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,62 @@
...
+                monolog-versions: [ '1.*', '2.*' ]
+                include:
+                    -   php-versions: '7.1'
+                    -   php-versions: '7.1'
+                        composer-flags: '--prefer-lowest'
+                    -   php-versions: '8.1'
+                        minimum-stability: 'dev'
+                    -   php-versions: 'nightly'
+                        minimum-stability: 'dev'
+        continue-on-error: ${{ matrix.php-versions > '8.0' }}
```
stof, line 24: Continuing after a step has failed looks weird to me. I suspect that you used this as an equivalent of the `allow_failures` feature of Travis, but this is absolutely not an equivalent feature (and I haven't found a case where it is the right solution anyway)

derrabus:
> Running tests on PHP 8.1 requires using PHPUnit 9.5 rather than 9.4 to avoid the deprecations

symfony/symfony#43200

.github/workflows/ci.yaml
HypeMC: Done

.github/workflows/ci.yaml
HypeMC: The default value is [true](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast), I've set it to false now.

.github/workflows/ci.yaml
HypeMC: Removed

HypeMC:
@stof @derrabus The deprecation is still present on PHPUnit 9.5 https://github.com/HypeMC/monolog-bundle/actions/runs/1280019019 . Perhaps the [baseline feature](https://symfony.com/doc/current/components/phpunit_bridge.html#baseline-deprecations) could be used?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,65 @@
...
+    pull_request:
+
+jobs:
+    test:
+        runs-on: ubuntu-latest
+        name: PHPUnit Tests
+        strategy:
+            fail-fast: false
+            matrix:
+                php-versions: [ '7.2', '7.3', '7.4', '8.0', '8.1', 'nightly' ]
```
derrabus, line 14:
```suggestion
                php: [ '7.2', '7.3', '7.4', '8.0' ]
```

Let's reduce the verbosity of the matrix variables a bit.

And let's limit the CI to the stable PHP versions for now. We can add PHP 8.1 and 8.2 in a follow-up if we like.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,65 @@
...
+                monolog-versions: [ '1.*', '2.*' ]
```
derrabus, line 15:
```suggestion
                monolog: [ '1.*', '2.*' ]
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,65 @@
...
+            -   name: Cache dependencies
+                uses: actions/cache@v2
+                with:
+                    path: ${{ steps.composer-cache.outputs.dir }}
+                    key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.json') }}
+                    restore-keys: ${{ runner.os }}-composer-
+
+            -   name: Configure composer minimum stability
+                if: ${{ matrix.minimum-stability != '' }}
+                run: composer config minimum-stability ${{ matrix.minimum-stability }}
```
derrabus, line 48:
```suggestion
            -   name: Configure composer
                if: "${{ matrix.deps == 'highest' }}"
                run: composer config minimum-stability dev
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,65 @@
...
+                php-versions: [ '7.2', '7.3', '7.4', '8.0', '8.1', 'nightly' ]
+                monolog-versions: [ '1.*', '2.*' ]
+                include:
+                    -   php-versions: '7.1'
+                    -   php-versions: '7.1'
+                        composer-flags: '--prefer-lowest'
+                    -   php-versions: '8.1'
+                        minimum-stability: 'dev'
+                    -   php-versions: 'nightly'
+                        minimum-stability: 'dev'
```
derrabus, line 23:
```suggestion
                include:
                    -   php: '7.4'
                        deps: lowest
                        deprecations: max[self]=0
                    -   php: '8.0'
                        deps: highest
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,65 @@
...
+                    key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.json') }}
+                    restore-keys: ${{ runner.os }}-composer-
+
+            -   name: Configure composer minimum stability
+                if: ${{ matrix.minimum-stability != '' }}
+                run: composer config minimum-stability ${{ matrix.minimum-stability }}
+
+            -   name: Require Monolog version
+                if: ${{ matrix.monolog-versions != '' }}
+                run: composer require --no-update monolog/monolog:${{ matrix.monolog-versions }}
```
derrabus, line 52:
```suggestion
                if: ${{ matrix.monolog != '' }}
                run: composer require --no-update monolog/monolog:${{ matrix.monolog }}
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,65 @@
...
+
+            -   name: Configure PHPUnit version
+                if: ${{ matrix.php-versions >= '8.1' }}
+                run: echo "SYMFONY_PHPUNIT_VERSION=9.5" >> $GITHUB_ENV
+
+            -   name: Install PHPUnit dependencies
+                run: vendor/bin/simple-phpunit install
+
+            -   name: Run tests
+                run: vendor/bin/simple-phpunit -v --coverage-text
```
derrabus, line 65:
```suggestion
                run: vendor/bin/simple-phpunit -v --coverage-text
                env:
                    SYMFONY_DEPRECATIONS_HELPER: '${{ matrix.deprecations }}'
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,65 @@
...
+
+            -   name: Install dependencies
+                run: composer update ${{ matrix.composer-flags }}
+
+            -   name: Configure PHPUnit version
+                if: ${{ matrix.php-versions >= '8.1' }}
+                run: echo "SYMFONY_PHPUNIT_VERSION=9.5" >> $GITHUB_ENV
+
+            -   name: Install PHPUnit dependencies
+                run: vendor/bin/simple-phpunit install
```
derrabus, line 62:
```suggestion
            -   name: Composer install
                uses: ramsey/composer-install@v1
                with:
                    dependency-versions: '${{ matrix.deps }}'
```

derrabus (changes requested):
I've streamlined the config a bit with the one in https://github.com/symfony/security-acl. I hope, that's okay. 🙂

.github/workflows/ci.yaml
HypeMC: Done

.github/workflows/ci.yaml
HypeMC: Done

.github/workflows/ci.yaml
HypeMC: @derrabus Unfortunately, PHP 8.0 + minimum-stability dev also fail cause of the same deprecation https://github.com/HypeMC/monolog-bundle/runs/3736881743

.github/workflows/ci.yaml
HypeMC: @derrabus Done, was PHP 7.1 removed intentionally?

.github/workflows/ci.yaml
HypeMC: Done

.github/workflows/ci.yaml
HypeMC: Done

.github/workflows/ci.yaml
HypeMC: Done

.github/workflows/ci.yaml
derrabus: Oh no, please re-add it. Sorry.

.github/workflows/ci.yaml
derrabus: Very well, no high-deps for now.

derrabus (approved):

.github/workflows/ci.yaml
HypeMC: Done

derrabus:
Thank you @HypeMC.

derrabus:
Thank you very much. If you want to work on a follow-up for PHP 8.1 and high-deps jobs, please go ahead. GitHub Actions should now be enabled on PRs thanks to your work.

PR #421, opened 2021-11-04
Add priority field to processor tags

sudo-plz:
So that the order in which the processors are used by monolog can be determined.

Example config:

`
    Foo\Bar\Log\Processors\FirstProcessor:
        tags:
            - { name: monolog.processor, channel: http }

    Foo\Bar\Log\Processors\SecondProcessor:
        tags:
            - { name: monolog.processor, channel: http, priority: -10 } # last

    Foo\Bar\Log\Processors\ThirdProcessor:
        tags:
            - { name: monolog.processor, channel: http, priority: -20 } # very last`

### DIFF
--- a/DependencyInjection/Compiler/AddProcessorsPass.php
+++ b/DependencyInjection/Compiler/AddProcessorsPass.php
@@ -28,32 +28,57 @@ class AddProcessorsPass implements CompilerPassInterface
             return;
         }
 
+        $processors = [];
+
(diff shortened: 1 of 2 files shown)


GromNaN (approved):
Thanks proposing this new feature. I think can be useful in rare cases the processors are dependent on each other.

It needs to be rebased on master. Features of PHP 7.1+ can be used.

PR #428, opened 2022-02-11
Leverage AsMonologProcessor attribute

fancyweb:
This feature allows configuring Monolog processors with an attribute. I basically copied `AsEventListener`.

### DIFF
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Bundle\MonologBundle\DependencyInjection;
 
+use Monolog\Attribute\AsMonologProcessor;
 use Monolog\Handler\FingersCrossed\ErrorLevelActivationStrategy;
 use Monolog\Logger;
 use Monolog\Processor\ProcessorInterface;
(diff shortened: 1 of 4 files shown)


derrabus:
Shall we add the attribute to the Monolog Bridge instead?

derrabus:
> Since the tag and its attributes are only consumed by the bundle, I thought it belonged here.

The same is true for any other `As*` attribute. They're all just "consumed" by FrameworkBundle, yet they reside inside the individial components. I'd place this attribute in the bridge and leave the integration logic in the bundle.

fancyweb:
Yes but this bundle is outside symfony/symfony, doesn't it make it different?

Adding the attribute in symfony/symfony would have to be done on 6.1 as a new feature I guess. Therefore, the whole feature would only be available once 6.1 is out and only with symfony/monolog-bridge >= 6.1 :sweat_smile: If everything is done here, it can be released earlier and it will be available with symfony/monolog-bridge >= 5.4 :man_shrugging:

nicolas-grekas:
I think we should submit the attribute to monolog!

When the attribute is set, we should also configure the binding. Here is where it's done for the corresponding marker interface.
We should preserve this behavior:

https://github.com/symfony/symfony/blob/44db49fb8f6d70c4341120a3f8f1b1a1a0dfa014/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Compiler/RegisterTokenUsageTrackingPass.php#L39-L42

fancyweb:
Thanks @derrabus and @nicolas-grekas for your explanations.

@derrabus, (and maybe @Seldaek), WDYT of adding the attribute directly in Monolog?

derrabus:
If the attribute is added to Monolog, great. If not, I'm still in favor of adding it to the bridge. 🙂

Seldaek:
I'm not sure how much sense this makes in Monolog though as it seems really specific to the DI autowiring here, but on the other hand it's a tiny zero-maintenance piece of code so I'm also not strictly opposed to it.

The channel/handler are optional is that correct? And method is only needed if the attribute is set to a class?

Seldaek:
Oh right of course, well I guess send a PR, but please add some docs explaining the intent of each property, and a class level docblock should explain what this is meant for but that this offers no guarantee that a processor is seen as such when using Monolog alone. Or maybe we could support it too in addProcessor if the class does not implement ProcessorInterface try to read the attribute to see if a method property is defined? Anyway let's see the details on the monolog repo.

Seldaek:
Thanks

Issue #412, opened 2021-07-04
"child process 5920 exited with status 3221225725 -- Restarting" & "The connection was reset (net::ERR_CONNECTION_RESET)" when using monolog on XAMPP


JPustkuchen:
### Context:
- XAMPP on Windows
- PHP 8.0.3

### Problem:
Monolog seems to eat up ressources (ThreadStackSize? ThreadsPerChild? MaxRequestsPerChild? - reasons unknown yet) which leads to a mysterious problem on Windows / WAMP. In my case it's XAMPP, but I guess it might also happen on other machines as they might use the same defaults.

Using Symfony with monolog (e.g. simply reproducable with https://github.com/symfony/demo) leads to these horrible errors:
**Browser:**
"The connection was reset" net::ERR_CONNECTION_RESET
![grafik](https://user-images.githubusercontent.com/637807/124377131-b4c08900-dcaa-11eb-8b8e-f4a4d80409fb.png)

**errorlog:**
[Fri Jul 02 06:42:45.297942 2021] [mpm_winnt:notice] [pid 5920:tid 684] AH00354: Child: Starting 150 worker threads.
[Fri Jul 02 06:42:57.561559 2021] [mpm_winnt:notice] [pid 4312:tid 652] AH00428: Parent: child process 5920 exited with status 3221225725 -- Restarting.

After two days and nights ( ;) ) I found out that it was monolog causing the problem. **As I removed** 
`Symfony\Bundle\MonologBundle\MonologBundle::class => ['all' => true],`
and the _monolog.yaml_ **everything simply works on WAMP**! Side loads, no errors!

My first thought was, that this might be caused by an infinite loop caused by an Error / exception but there's definitely no Error / Exception to be handled by Monolog and even it that had been the case, monolog should not fail and end up in such an inexpressive nightmare error like this.
We all know these kinds of errors where you have no clue where it may come from. This is one.

So from my perspective you can discuss if the related limits (see https://stackoverflow.com/a/43120319/10031165) are too low by default. But there's no reason to me, why monolog should eat up resources like this, especially if there isn't even an error.

Also see
- https://stackoverflow.com/a/43120319/10031165
- https://serverfault.com/questions/105908/how-do-you-increase-the-apache-connection-limit-wamp/105952
- https://stackoverflow.com/questions/5058845/how-do-i-increase-the-stack-size-for-apache-running-under-windows-7

[...]

PR #432, opened 2022-04-21
Split configuration

raziel057:
For better readability and to fix https://github.com/symfony/monolog-bundle/issues/412
As proposed here https://github.com/symfony/monolog-bundle/issues/412#issuecomment-1091780515

Workarround for PHP issue php/php-src#8315

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -356,7 +356,7 @@ class Configuration implements ConfigurationInterface
         $treeBuilder = new TreeBuilder('monolog');
         $rootNode = method_exists(TreeBuilder::class, 'getRootNode') ? $treeBuilder->getRootNode() : $treeBuilder->root('monolog');
 
-        $handlerNode = $rootNode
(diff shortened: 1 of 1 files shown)


raziel057:
Failing tests for PHPUnit Test for PHP < 7.4 is related to

raziel057:
Hi there, it seems that the ticket on PHP side (marked as Feature https://github.com/php/php-src/issues/8315) will not be resolved soon I propose to accept the following PR where the config is split in multiple sub parts to avoid to encounter the PHP issue.

ahajzik-crido:
Could you process this PR? Even though we have production env on linux, we develop on windows. It is hard to work doing "cache:clear" almost every time we need to test changes. We can't live long time with this issue. If it si not going to be solved we will be forced to abandon monolog logger.

Seldaek (approved):

Seldaek:
Thanks - kinda hard to review all this code shifting, but I'm hoping you did it diligently :)

PR #434, opened 2022-04-26
Fix potentials handlers added multiple times

### LINKED ISSUE #433: Logs logged multiple times (twice) in dev env due to redefining channels in handlers configuration
Hi,

I was since a long time having some logs logged twice on dev, I just figured out why but I'm not sure of the right way to fix it.

here is my handlers configuration:
```
# config/packages/monolog.yaml
monolog:
    handlers:
        worker:
            type: rotating_file
            max_files: 7
            level: !php/const Psr\Log\LogLevel::INFO
            path: '%kernel.logs_dir%/%kernel.environment%/worker.log'
            channels: [ worker ]
            bubble: true
```

```
# config/packages/dev/monolog.yaml
monolog:
    handlers:
        worker:
            type: rotating_file
            max_files: 2
            level: !php/const Psr\Log\LogLevel::DEBUG
            path: '%kernel.logs_dir%/%kernel.environment%/worker.log'
            channels: [worker]
            bubble: true
```

The fact that I'm redefining channels (even if identical) make the worker handler being added twice in LoggerChannelPass (pushHandler two times). I think we can fix it by:
- not redefining identical channels...
- using a array_unique Symfony\Bundle\MonologBundle\DependencyInjection\Configuration when validating channels

I'll try submiting a Pull Request but feel free to tell me if you think about an other (better) way to fix it

axi:
This seems to fix #433 Issue

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -781,6 +781,7 @@ class Configuration implements ConfigurationInterface
                                             return null;
                                         }
 
+                                        // de-duplicating $elements here in case the handlers are redefined, see https://github.com/symfony/monolog-bundle/issues/433
                                         return ['type' => $isExclusive ? 'exclusive' : 'inclusive', 'elements' => array_unique($elements)];
                                     })
                                 ->end()
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -781,7 +781,7 @@ class Configuration implements ConfigurationInterface
                                             return null;
                                         }
 
-                                        return ['type' => $isExclusive ? 'exclusive' : 'inclusive', 'elements' => $elements];
+                                        return ['type' => $isExclusive ? 'exclusive' : 'inclusive', 'elements' => array_unique($elements)];
                                     })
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks for the fix - the root cause here though IMO is that you should not be redefining the logging config at all, you should define it differently in every env, something like:

```yaml
when@dev:
    monolog:
        handlers:
            main:
                type:  stream
                path:  '%kernel.logs_dir%/%kernel.environment%.log'
                level: debug
                channels: ['!event']

when@prod:
    monolog:
        handlers:
            main:
                type: fingers_crossed
                action_level: error
                handler: nested
                excluded_http_codes: [404, 405]
                buffer_size: 100 # How many messages should be saved? Prevent memory leaks
            nested:
                type: stream
                path: "%kernel.logs_dir%/%kernel.environment%.log"
                level: debug
                include_stacktraces: true
```

So every env gets its own definition - and you avoid any config merging bugs.

axi:
I ended up also cleaning my config but hopefully, this will prevent somebody to have to search why he sees multiple logging lines in his log files :-)

PR #436, opened 2022-04-29
Add fallbackgroup handler to configuration documentation

mdeboer:
The fallbackgroup is already supported but wasn't listed in the configuration documentation.


Seldaek:
Thanks

PR #437, opened 2022-05-09
Allow installing Monolog 3, and run test suite with it

Seldaek:


.github/workflows/ci.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
             fail-fast: false
             matrix:
                 php: [ '7.2', '7.3', '7.4', '8.0', '8.1' ]
-                monolog: [ '1.*', '2.*' ]
+                monolog: [ '1.*', '2.*', '3.*@dev' ]
```
derrabus, line 15: This matrix spawns jobs with PHP 7.2 and Monolog 3 which is pointless.

Instead, let's add an entry to the `include` section:

```YAML
-   php: '8.1'
    monolog: '3.x@dev'
```

Seldaek: Yeah.. this was a rush job and it shows 😅 Thanks

Tests/DependencyInjection/ConfigurationTest.php
```diff
@@ -266,7 +270,7 @@ public function testWithConsoleHandler()
         $this->assertSame([
             OutputInterface::VERBOSITY_NORMAL => Logger::NOTICE,
             OutputInterface::VERBOSITY_VERBOSE => Logger::INFO,
-            OutputInterface::VERBOSITY_VERY_VERBOSE => 150,
+            OutputInterface::VERBOSITY_VERY_VERBOSE => 200,
```
stof, line 273: shouldn't this use a constant ?

Seldaek: Yes it should but I assume it was testing explicitly with a string value as technically people can use those in the config files. Anyway I just fixed it to have a valid level as 150 does not exist.

Seldaek:
Note that the 1.*/2.* failures are not related to this PR I believe - seems to be some deprecation warnings from symfony/monolog-bridge 5.4

derrabus (approved):

PR #438, opened 2022-05-10
Deprecate elasticsearch type, and split it into elastica and elastic_search types

### LINKED ISSUE #424: Collision between ElasticSearchHandler and ElasticaHandler
Hi,

I'm trying to set up monolog-bundle with native Elasticsearch cliente (introduced in monolog 2.0).
But it seems that I'm doing something wrong and I can't make it work.

Reviewing the code itself I've seen the following lines in MonologExtension:

https://github.com/symfony/monolog-bundle/blob/fde12fc628162787a4e53877abadc30047fd868b/DependencyInjection/MonologExtension.php#L983-L1011

There you can see that "elasticsearch" is always replaced with the ElasticaHandler in the array merge on line 1006.
Apart from that, I think the elastic handler should be ElasticsearchHandler, not ElasticSearchHandler (note te captial "S").

I locally did those changes but couldn't make it work and I'm lost.
Does symfony/monolog-bundle support the native Elasticsearch client? Or am I doing something wrong?

Many many thanks for your help

Seldaek:
Fixes #424

### DIFF
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -305,38 +305,51 @@ class MonologExtension extends Extension
             break;
 
         case 'elasticsearch':
-            if (isset($handler['elasticsearch']['id'])) {
-                $elasticaClient = new Reference($handler['elasticsearch']['id']);
-            } else {
-                // elastica client new definition
-                $elasticaClient = new Definition('Elastica\Client');
-                $elasticaClientArguments = [
-                    'host' => $handler['elasticsearch']['host'],
-                    'port' => $handler['elasticsearch']['port'],
-                    'transport' => $handler['elasticsearch']['transport'],
-                ];
+        case 'elastica':
+        case 'elastic_search':
+            if ($handler['type'] === 'elasticsearch') {
+                @trigger_error('The "elasticsearch" handler type is deprecated in MonologBundle since version 3.8.0, use the "elastica" type instead, or switch to the official Elastic client using the "elastic_search" type.', E_USER_DEPRECATED);
+            }
 
-                if (isset($handler['elasticsearch']['user'], $handler['elasticsearch']['password'])) {
-                    $elasticaClientArguments = array_merge(
-                        $elasticaClientArguments,
-                        [
-                            'headers' => [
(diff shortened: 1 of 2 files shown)


DependencyInjection/MonologExtension.php
```diff
@@ -305,38 +305,51 @@ private function buildHandler(ContainerBuilder $container, $name, array $handler
             break;
 
         case 'elasticsearch':
+        case 'elastica':
+        case 'elastic_search':
+            if ($handler['type'] === 'elasticsearch') {
+                @trigger_error('The "elasticsearch" handler type is deprecated in MonologBundle since version 3.8.0, use the "elastica" type instead, or switch to the official Elastic client using the "elastic_search" type.', E_USER_DEPRECATED);
+            }
+
```
derrabus, line 313:
```suggestion
            @trigger_error('The "elasticsearch" handler type is deprecated in MonologBundle since version 3.8.0, use the "elastica" type instead, or switch to the official Elastic client using the "elastic_search" type.', E_USER_DEPRECATED);
            // no break

        case 'elastica':
        case 'elastic_search':
```

derrabus (commented):
Should we add tests for this change?

stof (approved):

Seldaek:
See https://github.com/symfony/monolog-bundle/commit/769b7daa4f7a350d665c9334a2c0c71a683ea7c6 for a last fix here, forgot to add the factory

Chameleon-m:
elastic_search.host does not work, must be array
https://github.com/symfony/monolog-bundle/blob/master/DependencyInjection/MonologExtension.php#L292
need array for elastic_search, and key "hosts", example:
`
$clientArguments = [
    'hosts' => $handler['elasticsearch']['host'],
];
`
OR
`
$clientArguments = [
    'hosts' => $handler['elasticsearch']['hosts'],
];
`
And fix Configuration, host must be array|string:
https://github.com/symfony/monolog-bundle/blob/master/DependencyInjection/Configuration.php#L98
https://github.com/symfony/monolog-bundle/blob/master/DependencyInjection/Configuration.php#L879

JulienMarliac:
Hey

I can confirm that elastic_search does not work (see @Chameleon-m message) , any plans on the fix ? Or should I open a new issue ?

Kenneth-Sills:
[Ticket created.](https://github.com/symfony/monolog-bundle/issues/500)

PR #439, opened 2022-05-21
Allow using `OutputInterface::VERBOSITY_*` constants in PHP config

HeahDude:
While trying to improve some configuration blocks in Symfony docs, I found out that it was not possible to use constants for the verbosity levels of the console handler.
I think it's missing now that we tend to spread PHP config files such as `config/packages/monolog.php`.

Before this patch one gets the following exception:
```
Symfony\Component\Config\Definition\Exception\InvalidConfigurationException: Unrecognized options "32, 128" under "monolog.handlers.console.verbosity_levels". Available options are "VERBOSITY_DEBUG", "VERBOSITY_NORMAL", "VERBOSITY_QUIET", "VERBOSITY_VERBOSE", "VERBOSITY_VERY_VERBOSE".
```

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -17,6 +17,7 @@ use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
 use Symfony\Component\Config\Definition\ConfigurationInterface;
 use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
+use Symfony\Component\Console\Output\OutputInterface;
 
 /**
  * This class contains the configuration information for the bundle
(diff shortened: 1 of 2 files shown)


DependencyInjection/Configuration.php
```diff
@@ -1015,10 +1016,13 @@ private function addVerbosityLevelSection(ArrayNodeDefinition $handerNode)
                         ->then(function ($v) {
                             $map = [];
                             $verbosities = ['VERBOSITY_QUIET', 'VERBOSITY_NORMAL', 'VERBOSITY_VERBOSE', 'VERBOSITY_VERY_VERBOSE', 'VERBOSITY_DEBUG'];
-                            // allow numeric indexed array with ascendning verbosity and lowercase names of the constants
+                            $verbosityConstants = [OutputInterface::VERBOSITY_QUIET, OutputInterface::VERBOSITY_NORMAL, OutputInterface::VERBOSITY_VERBOSE, OutputInterface::VERBOSITY_VERY_VERBOSE, OutputInterface::VERBOSITY_DEBUG];
```
GromNaN, line 1019: The array should look like `[OutputInterface::VERBOSITY_QUIET => 'VERBOSITY_QUIET', ...]` to remove the `array_search` and make the code more readable.

GromNaN (commented):
👍🏻 for allowing constant instead of "magic" strings.

There is a 2nd transformation of constant name to constant value in the  config validation callback. All that may be simplified by converting to const values in the normalization.

The only drawback is that there will be int displayed instead of constant names when using `debug:config`.

DependencyInjection/Configuration.php
GromNaN: I think we can make the code a little clearer with this changes: https://github.com/HeahDude/monolog-bundle/pull/1

GromNaN:
Hi @HeahDude, do you want to rebase this PR after merging https://github.com/HeahDude/monolog-bundle/pull/1 (if you agree with its changes of-course)?

GromNaN (approved):

GromNaN:
Thank you @HeahDude.

PR #446, opened 2022-11-26
Resolve "level" env var at runtime

GromNaN:
Resolves #413, #444

Currently, the log level is read when the container built. This means we cannot change a log level at runtime using env vars.

With this change, log level can be modified for a specific command run: `LOG_LEVEL=debug bin/console ...`

~I would like to get some feedbacks before updating tests. I don't know yet what is the best strategy to check the resolution of the values.~

PR #448, opened 2022-12-19
Allow to use env var for log levels of console logger

GromNaN:
Following https://github.com/symfony/monolog-bundle/pull/446

The console logger have a map of verbosity => log level that should be configurable with env vars.

Disadvantage: in case of invalid log level, the exception cannot be contextualized with the output verbosity.

The tests in this PR will conflict with https://github.com/symfony/monolog-bundle/pull/439

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -1047,19 +1047,6 @@ class Configuration implements ConfigurationInterface
                                     ));
                                 }
 
-                                try {
-                                    if (Logger::API === 3) {
-                                        $level = Logger::toMonologLevel($level)->value;
-                                    } else {
-                                        $level = Logger::toMonologLevel(is_numeric($level) ? (int) $level : $level);
-                                    }
-                                } catch (\Psr\Log\InvalidArgumentException $e) {
-                                    throw new InvalidConfigurationException(sprintf(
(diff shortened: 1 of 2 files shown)


DependencyInjection/Configuration.php
```diff
@@ -1047,19 +1047,6 @@ private function addVerbosityLevelSection(ArrayNodeDefinition $handerNode)
...
                                 }
 
-                                try {
-                                    if (Logger::API === 3) {
-                                        $level = Logger::toMonologLevel($level)->value;
-                                    } else {
-                                        $level = Logger::toMonologLevel(is_numeric($level) ? (int) $level : $level);
-                                    }
-                                } catch (\Psr\Log\InvalidArgumentException $e) {
-                                    throw new InvalidConfigurationException(sprintf(
```
GromNaN, line 1057: To keep this validation, we could evaluate the value at build time, but pass the configured value (with env placeholder).

https://github.com/symfony/monolog-bundle/pull/446/files#diff-3ee12206c3b122be4593dbfdb19397c16b3d337f759fad3bde6c785d6f1261c0L50-L75

lyrixx (approved):

GromNaN:
Closing as this this removes the validation during the build time. The new `only_interactive` option would solve the initial issue I had with having a "console" only when running the command manually.

PR #454, opened 2023-03-04
Add `handler_default_channels` configuration option

HypeMC:
When configuring multiple handlers, one annoying problem I often come across is having to add every channel that is meant to be used with a specific handler to all handlers with an exclusive list, eg:

```yaml
monolog:
    handlers:
        main:
            type: fingers_crossed
            action_level: error
            handler: nested
            channels: [ "!one", "!two", "!three" ]
        nested:
            type: stream
            path: php://stderr
            level: debug
            formatter: monolog.formatter.json
        console:
            type: console
            process_psr_3_messages: false
            channels: [ "!event", "!doctrine", "!one", "!two", "!three" ]
```

The idea behind this PR is to be able to set a list of default channels that is applied to all handlers.

```diff
monolog:
+   handler_default_channels: [ "!one", "!two", "!three" ]
    handlers:
        main:
            type: fingers_crossed
            action_level: error
            handler: nested
-           channels: [ "!one", "!two", "!three" ]
        nested:
            type: stream
            path: php://stderr
            level: debug
            formatter: monolog.formatter.json
        console:
            type: console
            process_psr_3_messages: false
-           channels: [ "!event", "!doctrine", "!one", "!two", "!three" ]
+           channels: [ "!event", "!doctrine" ]
```

If a handler has a list of its own and it's the same type as the default one (inclusive/exclusive) the two lists are merged. If the handler's list is a different type, it replaces the default one. Handlers without lists inherit the default one. A `use_default_channels` handler option was also added so that a certain handler can choose not to use the default list.

   ```yaml
    monolog:
        handler_default_channels: [ "!one", "!two", "!three" ]
        handlers:
            one:
                # ...
                channels: [ "one", "four" ] # final list: ["one", "four"]
            two:
                # ...
                use_default_channels: false
                channels: [ "one", "four" ] # final list: ["one", "four"]
            three:
                # ...
                channels: [ "!two", "!three", "!four" ] # final list: ["!one", "!two", "!three", "!four"]
            four:
                # ...
                use_default_channels: false
                channels: [ "!two", "!three", "!four" ] # final list: ["!two", "!three", "!four"]
            five:
                # ...
                # final list: ["!one", "!two", "!three"]
   ```
   ```yaml
    monolog:
        handler_default_channels: [ "one", "two", "three" ]
        handlers:
            one:
                # ...
                channels: [ "one", "four" ] # final list: ["one", "two", "three", "four"]
            two:
                # ...
                use_default_channels: false
                channels: [ "one", "four" ] # final list: ["one", "four"]
            three:
                # ...
                channels: [ "!two", "!three", "!four" ] # final list: ["!two", "!three", "!four"]
            four:
                # ...
                use_default_channels: false
                channels: [ "!two", "!three", "!four" ] # final list: ["!two", "!three", "!four"]
            five:
            # ...
            # final list: ["one", "two", "three"]
   ```

### DIFF
--- a/config/schema/monolog-1.0.xsd
+++ b/config/schema/monolog-1.0.xsd
@@ -11,6 +11,7 @@
         <xsd:choice minOccurs="0" maxOccurs="unbounded">
             <xsd:element name="handler" type="handler" />
             <xsd:element name="channel" type="xsd:string" />
+            <xsd:element name="handler-default-channels" type="channels" minOccurs="0" maxOccurs="1" />
         </xsd:choice>
     </xsd:complexType>
 
@@ -93,6 +94,7 @@
         <xsd:attribute name="webhook-url" type="xsd:string" />
         <xsd:attribute name="slack-team" type="xsd:string" />
         <xsd:attribute name="region" type="xsd:string" />
+        <xsd:attribute name="use-default-channels" type="xsd:boolean" />
     </xsd:complexType>
 
     <xsd:simpleType name="level">
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -390,6 +390,8 @@ class Configuration implements ConfigurationInterface
         $treeBuilder = new TreeBuilder('monolog');
         $rootNode = $treeBuilder->getRootNode();
 
+        $this->addChannelsSection($rootNode, 'handler_default_channels');
+
         $handlers = $rootNode
             ->fixXmlConfig('channel')
             ->fixXmlConfig('handler')
@@ -634,6 +636,7 @@ class Configuration implements ConfigurationInterface
                 ->end()
                 ->scalarNode('formatter')->end()
                 ->booleanNode('nested')->defaultFalse()->end()
+                ->booleanNode('use_default_channels')->defaultTrue()->end()
             ->end();
 
         $this->addGelfSection($handlerNode);
@@ -1086,11 +1089,11 @@ class Configuration implements ConfigurationInterface
         ;
     }
 
-    private function addChannelsSection(ArrayNodeDefinition $handlerNode)
+    private function addChannelsSection(ArrayNodeDefinition $handlerNode, string $nodeName = 'channels')
     {
         $handlerNode
             ->children()
-                ->arrayNode('channels')
+                ->arrayNode($nodeName)
                     ->fixXmlConfig('channel', 'elements')
                     ->canBeUnset()
                     ->beforeNormalization()
@@ -1106,7 +1109,7 @@ class Configuration implements ConfigurationInterface
                         ->thenUnset()
                     ->end()
                     ->validate()
-                        ->always(function ($v) {
+                        ->always(function ($v) use ($nodeName) {
                             $isExclusive = null;
                             if (isset($v['type'])) {
                                 $isExclusive = 'exclusive' === $v['type'];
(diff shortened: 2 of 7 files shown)


Seldaek:
I kinda get the problem you're trying to solve but IMO the solution is a bit unclear. Are those auto excluded channels excluded from all handlers, or only those with exclusion lists? I think it's the latter but I don't really see why..

Also name nitpick, perhaps `default_excluded_channels` instead of auto. Having `default_channels` might be even more sensible so you can pick or exclude like the rest of `channels`, and it wouldn't be an inverted option.

HypeMC:
> I kinda get the problem you're trying to solve but IMO the solution is a bit unclear. Are those auto excluded channels excluded from all handlers, or only those with exclusion lists? I think it's the latter but I don't really see why..

Hi, they're excluded only from handlers with an exclusion list. I don't think it would make since for the channels to be excluded from inclusion lists since the channels in those lists are explicitly added through the configuration. I'm not sure how to make this a little more clear.

> Also name nitpick, perhaps `default_excluded_channels` instead of auto.

Your suggestion makes sense, I've updated the name.

> Having `default_channels` might be even more sensible so you can pick or exclude like the rest of `channels`, and it wouldn't be an inverted option.

Just to see if I've understood you correctly, the `default_channels` option would be a list of channels to be included to or excluded from every handler. If that's the idea, I like it a lot. How would this work in terms of precedence though? Here's my guess:

1. if `default_channels` is an exclusion list and the handlers `channels` list is an inclusion list, or, `default_channels` is an inclusion list and the handlers `channels` list is an exclusion list, the handlers list would take precedence:
   ```yaml
   monolog:
       default_channels: [ "!one", "!two", "!three" ]
       handlers:
           main:
               # ...
               channels: [ "one", "four" ] # final list: ["one", "four"]
   ```
   ```yaml
   monolog:
       default_channels: [ "one", "two", "three" ]
       handlers:
           main:
               # ...
               channels: [ "!one", "!four" ] # final list: ["!one", "!four"]
   ```
1. if both `default_channels` and the handlers `channels` list are exclusion lists, or, both are inclusion lists, the lists would get merged:
   ```yaml
   monolog:
       default_channels: [ "!one", "!two" ]
       handlers:
           main:
               # ...
               channels: [ "!two", "!three", "!four" ] # final list: ["!one", "!two", "!three", "!four"]
   ```
   ```yaml
   monolog:
       default_channels: [ "one", "two" ]
       handlers:
           main:
               # ...
               channels: [ "two", "three", "four" ] # final list: ["one", "two", "three", "four"]
   ```
In other words, if the lists are the same type, they'd get merged, otherwise, if they're not, the handlers `channels` list would take precedance.

Also, what about cases such as whitelisting a default excluded channel on some handler or blacklisting a default included one on another? Should there be some special syntax for that, eg my `!!` proposal:
```yaml
monolog:
    default_channels: [ "!one", "!two" ]
    handlers:
        main:
            # ...
            channels: [ "!!two", "!three", "!four" ] # final list: ["!one", "!three", "!four"]
```
or would adding a `use_default_channels` option make more sense. The option would be `true` by default, but could be set to `false` on a per-handler basis:
```yaml
monolog:
    default_channels: [ "!one", "!two" ]
    handlers:
        main:
            # ...
            use_default_channels: false
            channels: [ "!one", "!three", "!four" ] # final list: ["!one", "!three", "!four"]
```

@Seldaek Please let me know your thoughts on this and I'll update the PR accordingly.

HypeMC:
I've updated the PR according to my proposal.

GromNaN:
It seems complex to understand and perhaps a little too specific for something that can be solved with an `array_merge` using the PHP configuration format.

```php
<?php

use Symfony\Config\MonologConfig;

return static function (MonologConfig $monolog) {
    $defaultChannels = ['!one', '!two', '!three'];

    $monolog->handlers()->main()
        ->type('fingers_crossed')
        ->actionLevel('error')
        ->handler('nested')
        ->channels($defaultChannels);

    $monolog->handlers()->nested()
        ->type('stream')
        ->path('php://stderr')
        ->level('debug')
        ->formatter('monolog.formatter.json');

    $monolog->handlers()->console()
        ->type('console')
        ->processPsr3Messages(false)
        ->channels(['!event', '!doctrine', ...$defaultChannels]);
};
```

HypeMC:
@GromNaN You make a good point. This PR was opened with YAML in mind. While `array_merge()` works fine with PHP config, YAML doesn't have a similar option. While this approach works with all config formats, I agree it's a bit complex. I'm fine with closing it if you feel it's not worth it.

GromNaN:
The fact that we need a new `use_default_channels` option is a clear signal that it's too complex to apprehend. PHP configuration is a way to go.

Thank you for your understanding.

PR #461, opened 2023-08-22
Wire DebugLoggerConfigurator on Symfony 6.4+

nicolas-grekas:
Wire https://github.com/symfony/symfony/pull/51284 and allow Symfony v7, which will just work after https://github.com/symfony/symfony/pull/51458 is merged.


DependencyInjection/Configuration.php
```diff
@@ -375,10 +375,8 @@ class Configuration implements ConfigurationInterface
 {
     /**
      * Generates the configuration tree builder.
-     *
-     * @return \Symfony\Component\Config\Definition\Builder\TreeBuilder The tree builder
      */
-    public function getConfigTreeBuilder()
+    public function getConfigTreeBuilder(): TreeBuilder
```
nicolas-grekas, line 379: Technically this is a BC break for anyone that extends Configuration, but let's be pragmatic: nobody does that.

stof: Those configuration classes should be `@internal` anyway

stof (approved):

PR #464, opened 2023-09-14
Fix test for autoconfiguration

nkl-kst:
This PR contains two small fixes for the Monolog processor autoconfiguration test. The test currently fails because of changes in https://github.com/Seldaek/monolog/pull/1797.

Not sure if I should provide more information. Please let me know if I missed something 🙂

### DIFF
--- a/Tests/DependencyInjection/Fixtures/AsMonologProcessor/FooProcessorWithPriority.php
+++ b/Tests/DependencyInjection/Fixtures/AsMonologProcessor/FooProcessorWithPriority.php
@@ -1,0 +1,23 @@
+<?php
+
+/*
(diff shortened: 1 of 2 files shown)


nkl-kst:
After the original fix, there were failing tests for `monolog/monolog` v2 because the `AsMonologProcessor::priority` attribute doesn't exist there.

I decided to revert the original fix, added a new test for Monolog v3.4.0 and skip the existing and new tests conditionally based on the used Monolog version.

There's still one failing test setup (PHP 7.3 with Monolog 2.*), but I think this is not related to the changes in this PR.

derrabus (approved):

derrabus:
Thank you @nkl-kst.

PR #468, opened 2023-10-24
Add support for the `#[WithMonologChannel]` attribute of Monolog 3.5 to autoconfigure the channel

stof:
Configuring custom channels on some services is one of the last reasons I have for defining services explicitly in my config files. This attribute allows to solve that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased
 
+* Add support for the `WithMonologChannel` attribute of Monolog 3.5.0 to autoconfigure the `monolog.logger` tag
 * Add support for Symfony 7
 * Remove support for Symfony 4
 * Add support for env placeholders in the `level` option of handlers
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -12,6 +12,7 @@
 namespace Symfony\Bundle\MonologBundle\DependencyInjection;
 
 use Monolog\Attribute\AsMonologProcessor;
+use Monolog\Attribute\WithMonologChannel;
 use Monolog\Handler\FingersCrossed\ErrorLevelActivationStrategy;
 use Monolog\Handler\HandlerInterface;
 use Monolog\Logger;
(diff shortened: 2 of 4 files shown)


Attribute/WithLoggerChannel.php
```diff
@@ -0,0 +1,28 @@
...
+namespace Symfony\Bundle\MonologBundle\Attribute;
+
+#[\Attribute(\Attribute::TARGET_CLASS)]
+class WithLoggerChannel
+{
+    private string $channel;
+
+    public function __construct(string $channel)
+    {
+        $this->channel = $channel;
```
OskarStark, line 21: We can use CPP here, can we?

OskarStark: Oh no, we still support Symfony 5.4

stof: For a package supporting PHP 7.2+ ?

Attribute/WithLoggerChannel.php
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Autoconfigures the logger channel used when inject a logger in the service.
+ *
+ * This adds the `monolog.logger` tag on the service definition.
+ */
+#[\Attribute(\Attribute::TARGET_CLASS)]
+class WithLoggerChannel
+{
+    private string $channel;
```
nicolas-grekas, line 22: what about using a public property marked as `@readonly`?

stof: Done.

I also made the class final as there is no reason to extend it.

Attribute/WithLoggerChannel.php
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * This adds the `monolog.logger` tag on the service definition.
+ */
+#[\Attribute(\Attribute::TARGET_CLASS)]
+final class WithLoggerChannel
+{
+    /**
+     * @readonly
+     */
+    public string $channel;
```
nicolas-grekas, line 25:
```suggestion
    public $channel;
```

stof: IMO, it is fine to keep the typed property there. Attributes don't work on PHP 7.2 anyway

lyrixx (approved):
👍🏼 I like it

Attribute/WithLoggerChannel.php
stof: ah indeed. For some reasons, I thought CPP was 8.1 like the readonly properties.

nicolas-grekas:
> should we move the attribute to the bridge?

Or better: to monolog itself?

stof:
@Seldaek do you think this attribute should go in Monolog itself ?

stof:
I submitted it to Monolog in https://github.com/Seldaek/monolog/pull/1847
If this is accepted, I will update this PR to use the Monolog attribute instead.

PR #469, opened 2023-10-24
Tag classes as final or internal based on their usages

stof:
This will allow us to have a clearer BC API for our future versions.


GromNaN (approved):

OskarStark (approved):

DependencyInjection/Compiler/AddProcessorsPass.php
```diff
@@ -20,6 +20,8 @@
  * Registers processors in Monolog loggers or handlers.
  *
  * @author Christophe Coevoet <stof@notk.org>
+ *
+ * @internal since 3.8.0
```
derrabus, line 24: 3.8.0 has been released already hasn't it.

PR #470, opened 2023-10-31
Add support for additional parameters in native mailer

nkl-kst:
This PR adds support for additional parameters to the native mailer, as describe in https://www.php.net/manual/en/function.mail.php

> The additional_params parameter can be used to pass additional flags as command line options to the program configured to be used when sending mail, as defined by the sendmail_path configuration setting. For example, this can be used to set the envelope sender address when using sendmail with the -f sendmail option. 

In a recent project we had the requirement to set the additional parameter `-ftest@example.com` to be able to send emails through a relay server. Otherwise the mails were bounced.

The bundle currently allows to set these parameters only on existing logger instances. This PR allows to set them in the configuration file:

```yaml
native_mailer:
    type: native_mailer
    from_email: 'from@example.com'
    to_email: 'to@example.com'
    subject: 'Example subject'
    parameters:
        - '-ftest@example.com'
```

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -197,6 +197,7 @@ use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
  *   - [level]: level name or int value, defaults to DEBUG
  *   - [bubble]: bool, defaults to true
  *   - [headers]: optional array containing additional headers: ['Foo: Bar', '...']
+ *   - [parameters]: optional array containing additional parameters: ['--for bar', '...']
  *
  * - symfony_mailer:
  *   - from_email: optional if email_prototype is given
@@ -429,6 +430,7 @@ class Configuration implements ConfigurationInterface
                 ->fixXmlConfig('tag')
                 ->fixXmlConfig('accepted_level')
                 ->fixXmlConfig('header')
+                ->fixXmlConfig('parameter')
                 ->canBeUnset();
 
         $handlerNode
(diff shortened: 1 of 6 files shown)


GromNaN:
Thanks for this contribution! I rebased your branch onto the current `4.x` branch and retargeted the PR to `4.x`, since that is where new changes go now. The directory layout changed between 3.x and 4.x, so the files moved to `src/`, `config/` and `tests/`.

Summary of what changed:

- Added a `parameters` option to the `native_mailer` handler, mirroring the existing `headers` option. It maps to `NativeMailerHandler::addParameter()` and lets you pass extra parameters to `mail()`, e.g. `-f sender@example.com` to set the envelope sender.
- Added the matching XML config support (`fixXmlConfig('parameter')` and the `<parameter>` XSD element) so it works in XML too.
- Added a hardening on top of the original patch: unlike `addHeader()`, `addParameter()` performs no validation. Because these values end up on the sendmail command line through `mail()`, a value containing a newline could allow argument or header injection. Each parameter is now validated at config compile time and rejected if it contains `\r` or `\n`, mirroring the protection already in place for headers.
- Updated the `native_mailer` fixtures and test assertion, added a dedicated `ConfigurationTest` case for the newline rejection, and added a CHANGELOG entry.

Let me know if you want any adjustment.

GromNaN (approved):

GromNaN:
It took time, but here we go, this is in now. Thank you very much @nkl-kst.

PR #471, opened 2023-11-06
Add configuration support for SamplingHandler

Okhoshi:
Figured out the SamplingHandler was never added to the configuration, as per #4 to-do list :)


stof (approved):

stof:
@Okhoshi now :smile:

PR #472, opened 2023-11-09
Fix badly formatted internal/final annotations

norkunas:


GromNaN (approved):

OskarStark (approved):

stof (approved):

dennismetz:
Can we make a new release out of this?
We currently get the bug in 3.10.0, but don't want to use dev-master as a dependency.

Issue #441, opened 2022-07-22
Syslog formatter wiring


dkarlovi:
Monolog just merged Syslog formatter which should be wired here.
https://github.com/Seldaek/monolog/issues/1594

PR #473, opened 2024-01-19
Add service definition for SyslogFormatter

sarim:
| Q             | A
| ------------- | ---
| Branch?       | 3.x
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix https://github.com/symfony/monolog-bundle/issues/441
| License       | MIT

From https://github.com/symfony/recipes/issues/1111 & https://github.com/Seldaek/monolog/pull/1689 I found out about the (newish) syslog formatter that can be viewed nicely with lnav. While trying to use it, I see there is no service created for this in monolog-bundle. Users can obviously make their own service in their app's services.yaml, but seeing other formatters exposed and discussion about adding it via recipe, I thought it would be good idea to create the service from monolog-bundle.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * Add `enabled` option to `handlers` configuration
 * Add `priority` field to `processor` tag
 * Add `mongodb` handler and deprecate `mongo`
+* Add `monolog.formatter.syslog` service definition to format RFC5424-compliant messages
 
 ## 3.10.0 (2023-11-06)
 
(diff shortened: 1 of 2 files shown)


GromNaN:
I rebased the PR. I think we should merge it and unblock https://github.com/symfony/recipes/issues/1111

GromNaN (approved):

Seldaek (approved):
LGTM

GromNaN:
It took time, but here we go, this is in now. Thank you very much @sarim.

PR #482, opened 2024-05-27
Add enabled parameter in configuration

Arkalo2:
| Q             | A
| ------------- | ---
| Branch?       | 3.x
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Tickets       | 
| License       | MIT
| Doc PR        | 

This feature aims to separate the enabling of handlers based on the environment.

## Use Case :

You have 3 servers on your project : dev, staging, and prod
- On your **dev** machine, you are in the **dev** environment.
- On your **staging** server, you are in the **prod** environment.
- On your **prod** server, you are in the **prod** environment.

Currently, you can't disable handlers for a specific environment. For an instance, you can't disable logging only in your staging environment.

With the enabled parameter, you are able to disable logging only in the staging environment.

### DIFF
--- a/tests/DependencyInjection/Compiler/LoggerChannelPassTest.php
+++ b/tests/DependencyInjection/Compiler/LoggerChannelPassTest.php
@@ -14,6 +14,7 @@ namespace Symfony\Bundle\MonologBundle\Tests\DependencyInjection\Compiler;
 use PHPUnit\Framework\TestCase;
 use Psr\Log\LoggerInterface;
 use Symfony\Bundle\MonologBundle\DependencyInjection\Compiler\LoggerChannelPass;
+use Symfony\Bundle\MonologBundle\DependencyInjection\MonologExtension;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
@@ -231,6 +232,31 @@ class LoggerChannelPassTest extends TestCase
         $container->getCompilerPassConfig()->setRemovingPasses([]);
 
         return $container;
+    }
+
+    /**
+     * @testWith [true]
+     *           [false]
+     */
+    public function testEnabledHandler(bool $enabled)
+    {
+        $container = new ContainerBuilder();
+        $loader = new MonologExtension();
+
+        $config = [
+            'handlers' => [
+                'main' => [
+                    'enabled' => $enabled,
+                    'type' => 'stream',
+                    'path' => '%kernel.logs_dir%/%kernel.environment%.log',
+                    'level' => 'debug',
+                ],
+            ],
+        ];
+
+        $loader->load([$config], $container);
+
+        $this->assertSame($enabled, $container->hasDefinition('monolog.handler.main'));
     }
 }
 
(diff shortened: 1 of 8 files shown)


DependencyInjection/MonologExtension.php
```diff
@@ -75,6 +75,9 @@ public function load(array $configs, ContainerBuilder $container)
             $handlers = [];
 
             foreach ($config['handlers'] as $name => $handler) {
+                if(false === $handler['enabled']) {
```
nicolas-grekas, line 78: CS is borken
and if this is a boolean, there's no need for any comparisons:
```suggestion
                if (!$handler['enabled']) {
```

DependencyInjection/Configuration.php
```diff
@@ -446,6 +446,7 @@ public function getConfigTreeBuilder(): TreeBuilder
                     ->end()
                 ->end()
                 ->scalarNode('id')->end() // service & rollbar
+                ->scalarNode('enabled')->defaultTrue()->end()
```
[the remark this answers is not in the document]
GromNaN: Indeed, a `booleanNode` should be used here because this value is used during compilation and cannot accept an environment variable.

GromNaN (changes requested):
I find this feature useful. Would you like to rebase your PR on the branch 3.x (we moved files), and add a test case in [FixtureMonologExtensionTestCase](https://github.com/symfony/monolog-bundle/blob/3.x/tests/DependencyInjection/FixtureMonologExtensionTestCase.php)?

tests/DependencyInjection/Compiler/LoggerChannelPassTest.php
```diff
@@ -232,6 +233,48 @@ private function getFunctionalContainer()
...
+                    'type' => 'stream',
+                    'path' => '%kernel.logs_dir%/%kernel.environment%.log',
+                    'level' => 'debug',
+                ],
+            ],
+        ];
+
+        $loader->load([$config], $container);
+
+        $this->assertTrue($container->hasDefinition('monolog.handler.main'));
```
GromNaN: As the tests are almost identical, we can merge them by passing a boolean with a testWith annotation. That way we ensure there is no other reason that make definition non-existing.
```php
    /**
     * @testWith [true]
     *           [false]
     */
    public function testEnabledHandler($enabled): void
    {
        // ['enabled' => $enabled];

        $this->assertSame($enabled, $container->hasDefinition('monolog.handler.main'));
    }
```

Arkalo2: Cool ! I did not know

tests/DependencyInjection/Compiler/LoggerChannelPassTest.php
```diff
@@ -232,6 +233,52 @@ private function getFunctionalContainer()
...
+                ],
+            ],
+        ];
+
+        $loader->load([$config], $container);
+
+        $this->assertSame($enabled, $container->hasDefinition('monolog.handler.main'));
+    }
+
+    public function testDisabledHandler()
```
GromNaN: Then this test is a duplicate of the previous one with false

GromNaN (approved):

GromNaN:
Thank you @Arkalo2.

Arkalo2:
Thanks @GromNaN @nicolas-grekas for your reviews

I'll add a PR for documentation

PR #455, opened 2023-03-17
Add priority field to processor tags

sudo-plz:
So that the order in which the processors are used by monolog can be determined.
Example config:

```yaml
Foo\Bar\Log\Processors\FirstProcessor:
    tags:
        - { name: monolog.processor, channel: http }

Foo\Bar\Log\Processors\SecondProcessor:
    tags:
        - { name: monolog.processor, channel: http, priority: -10 } # last

Foo\Bar\Log\Processors\ThirdProcessor:
    tags:
        - { name: monolog.processor, channel: http, priority: -20 } # very last
```
(PR in favor of [old PR](https://github.com/symfony/monolog-bundle/pull/421))

PR #485, opened 2024-07-28
Add priority field to processor tags

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 3.x
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | -
| License       | MIT

Continuation of #455.

```yaml
Foo\Bar\Log\Processors\FirstProcessor:
    tags:
        - { name: monolog.processor, channel: http }

Foo\Bar\Log\Processors\SecondProcessor:
    tags:
        - { name: monolog.processor, channel: http, priority: -10 } # last

Foo\Bar\Log\Processors\ThirdProcessor:
    tags:
        - { name: monolog.processor, channel: http, priority: -20 } # very last
```

```php
#[AsMonologProcessor(priority: -10)]
class SecondProcessor {}

#[AsMonologProcessor(priority: -20)]
class ThirdProcessor {}
```

**Probably better to review with [whitespace changes ignored](https://github.com/symfony/monolog-bundle/pull/485/files?w=1).**

### DIFF
--- a/src/DependencyInjection/Compiler/AddProcessorsPass.php
+++ b/src/DependencyInjection/Compiler/AddProcessorsPass.php
@@ -11,10 +11,11 @@
 
 namespace Symfony\Bundle\MonologBundle\DependencyInjection\Compiler;
 
+use Symfony\Component\DependencyInjection\Argument\TaggedIteratorArgument;
 use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\Compiler\PriorityTaggedServiceTrait;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
-use Symfony\Component\DependencyInjection\Reference;
 
 /**
  * Registers processors in Monolog loggers or handlers.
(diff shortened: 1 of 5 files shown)


DependencyInjection/Compiler/AddProcessorsPass.php
```diff
@@ -25,16 +25,20 @@
...
+    use PriorityTaggedServiceTrait;
+
     public function process(ContainerBuilder $container)
     {
         if (!$container->hasDefinition('monolog.logger')) {
             return;
         }
 
-        foreach ($container->findTaggedServiceIds('monolog.processor') as $id => $tags) {
+        foreach (array_reverse($this->findAndSortTaggedServices('monolog.processor', $container)) as $reference) {
```
94noni, line 36: perhaps comment why `array_reverse` is used ?
otherwise its smart :)

94noni (approved):

DependencyInjection/Compiler/AddProcessorsPass.php
HypeMC: Good idea, done

DependencyInjection/Compiler/AddProcessorsPass.php
```diff
@@ -25,16 +25,21 @@
...
+
     public function process(ContainerBuilder $container)
     {
         if (!$container->hasDefinition('monolog.logger')) {
             return;
         }
 
-        foreach ($container->findTaggedServiceIds('monolog.processor') as $id => $tags) {
+        // array_reverse is used because ProcessableHandlerTrait::pushProcessor prepends processors to the beginning of the stack
+        foreach (array_reverse($this->findAndSortTaggedServices('monolog.processor', $container)) as $reference) {
```
stof, line 37: how does this behave when we have multiple tags (for instance when registering the processor in specific channels or in specific handlers ?

HypeMC: @stof Good catch, it didn't work. I've managed to fix it. Please let me know what you think of the fix. It's probably easiest to review it with [whitespace changes ignored](https://github.com/symfony/monolog-bundle/pull/485/commits/264cf5d4510096f149f0d29779c34f69ce3757c8?w=1).

derrabus:
@alexndlm Did you review and test the change in your projects? I agree that we should try to get it merged.

alexndlm:
Of course not, I have unlimited trust in Symfony 😎 But I will do this soon to double-check.

derrabus:
> But I will do this soon to double-check.

Thank you! I personally don't need this feature at the moment. So getting feedback from someone who'd actually use it is more than valuable to us.

nicolas-grekas:
Looks like PR is needed for https://github.com/symfony/symfony/pull/54445?

stof:
@nicolas-grekas no. this PR does not affect the Symfony PR (see my comment on the Symfony PR)

alexndlm:
@derrabus I have tested it, and it works perfectly

alexndlm:
@derrabus, my bad, it does not work; I have found one issue.

When we add some processor only for channel `app` with the lowest priority, it will always be first. This is because in `\Symfony\Bundle\MonologBundle\DependencyInjection\Compiler\AddProcessorsPass` for the `app` channel, we use `monolog.logger`, and when there is no channel, we use `monolog.logger_prototype`.

And in `symfony/monolog-bundle/Resources/config/monolog.xml` the `monolog.logger` has parent `monolog.logger_prototype`. 

So, when DI create `monolog.logger` service, it first calls push `pushProcessor` for `monolog.logger_prototype` and then calls for `monolog.logger`. 

The same issue should be for any channel.

Tests/DependencyInjection/Compiler/AddProcessorsPassTest.php
```diff
@@ -31,12 +32,28 @@ public function testHandlerProcessors()
...
-        $this->assertEquals(['pushProcessor', [new Reference('test')]], $calls[0]);
+        $this->assertEquals(['pushProcessor', [new TypedReference('test', 'TestClass')]], $calls[0]);
 
         $service = $container->getDefinition('handler_test');
         $calls = $service->getMethodCalls();
         $this->assertCount(1, $calls);
-        $this->assertEquals(['pushProcessor', [new Reference('test2')]], $calls[0]);
+        $this->assertEquals(['pushProcessor', [new TypedReference('test2', 'TestClass')]], $calls[0]);
+
+        $service = $container->getDefinition('monolog.handler.priority_test');
```
alexndlm, line 42: Can we add a case when the processor has a channel?

alexndlm (changes requested):

alexndlm:
@HypeMC, do you have time to fix this?

HypeMC:
> When we add some processor only for channel `app` with the lowest priority, it will always be first. This is because in `\Symfony\Bundle\MonologBundle\DependencyInjection\Compiler\AddProcessorsPass` for the `app` channel, we use `monolog.logger`, and when there is no channel, we use `monolog.logger_prototype`.
> 
> And in `symfony/monolog-bundle/Resources/config/monolog.xml` the `monolog.logger` has parent `monolog.logger_prototype`.
> 
> So, when DI create `monolog.logger` service, it first calls push `pushProcessor` for `monolog.logger_prototype` and then calls for `monolog.logger`.
> 
> The same issue should be for any channel.

@alexndlm Sorry for the late reply, I missed your feedback. The issue should be fixed with https://github.com/symfony/monolog-bundle/pull/485/commits/1ad6fc7ee4bdc3ea9289234bb3bcef3b79da6857.

Tests/DependencyInjection/Compiler/AddProcessorsPassTest.php
HypeMC: Done

src/DependencyInjection/Compiler/AddProcessorsPass.php
```diff
@@ -25,48 +26,78 @@
...
+                if (!method_exists($class, 'pushProcessor')) {
+                    throw new \InvalidArgumentException(\sprintf('The "%s" handler does not accept processors.', $tag['handler']));
+                }
+            } elseif (!empty($tag['channel'])) {
+                $loggerId = 'app' === $tag['channel'] ? 'monolog.logger' : \sprintf('monolog.logger.%s', $tag['channel']);
+                $definitions = [$container->getDefinition($loggerId)];
+            } else {
+                if ($this->channelPass) {
+                    $definitions = [];
+                    foreach ($this->channelPass->getChannels() as $channel) {
```
GromNaN: The compiler passes should be stateless. We can add a tag to the logger definitions so that we can retrieve them with `$container->findTaggedServiceIds('monolog.channel_logger')`

src/DependencyInjection/Compiler/AddProcessorsPass.php
```diff
@@ -25,48 +26,78 @@
...
+                }
+                $class = $container->getParameterBag()->resolveValue($parentDef->getClass());
+                if (!method_exists($class, 'pushProcessor')) {
+                    throw new \InvalidArgumentException(\sprintf('The "%s" handler does not accept processors.', $tag['handler']));
+                }
+            } elseif (!empty($tag['channel'])) {
+                $loggerId = 'app' === $tag['channel'] ? 'monolog.logger' : \sprintf('monolog.logger.%s', $tag['channel']);
+                $definitions = [$container->getDefinition($loggerId)];
+            } else {
+                if ($this->channelPass) {
```
GromNaN: Nit: you can combine `else` and `if` into `elseif`

GromNaN (approved):
LGTM, with some improvement that can be done later.

src/DependencyInjection/Compiler/AddProcessorsPass.php
HypeMC: Done, https://github.com/symfony/monolog-bundle/compare/62fae471255ae474c4ee85dc3b1a550178e06ae4..676eb84aec63cd0ceb01bb03aeca7f19712b12e5

src/DependencyInjection/Compiler/AddProcessorsPass.php
HypeMC: Good catch, done

GromNaN:
Thank you @HypeMC.

PR #486, opened 2024-07-28
Fix CI

HypeMC:
Fixes https://github.com/symfony/monolog-bundle/actions/runs/10134253734/job/28020275060

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
     "require-dev": {
         "symfony/yaml": "^5.4 || ^6.0 || ^7.0",
         "symfony/console": "^5.4 || ^6.0 || ^7.0",
-        "symfony/phpunit-bridge": "^6.3 || ^7.0"
+        "symfony/phpunit-bridge": "^7.1"
     },
     "autoload": {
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -26,7 +26,7 @@
     "require-dev": {
         "symfony/yaml": "^5.4 || ^6.0 || ^7.0",
         "symfony/console": "^5.4 || ^6.0 || ^7.0",
-        "symfony/phpunit-bridge": "^6.3 || ^7.0"
+        "symfony/phpunit-bridge": "^6.3.10 || ^7.0"
```
derrabus, line 29:
```suggestion
        "symfony/phpunit-bridge": "^7.1"
```

Testing with ancient versions of the PHPUnit bridge is pointless.

derrabus (changes requested):

composer.json
HypeMC: @derrabus Done

derrabus (approved):

derrabus:
Thank you @HypeMC.

PR #487, opened 2024-07-28
Add PHP-CS-Fixer

HypeMC:
This repo seems to be the only one that doesn't use PHP-CS-Fixer, which is a bit annoying when trying to contribute since I have to manually worry about coding standards. I'm not sure if there's a specific reason for that, so I decided to give it a try in this PR.

I took the `.php-cs-fixer.dist.php` from the framework repo and made a few adjustments. I removed `nullable_type_declaration` and `nullable_type_declaration_for_default_null_value` since those are part of the standard Symfony set. Additionally, I set `modernize_strpos` to `false` since the bundle still supports PHP >= 7.2.

I'm not sure how to run it in the CI, as some repos use fabbot while others, like the MakerBundle, use GitHub Actions.

It would be best to hide whitespace changes while reviewing.


DependencyInjection/Compiler/AddProcessorsPass.php
```diff
@@ -34,24 +34,24 @@ public function process(ContainerBuilder $container)
         foreach ($container->findTaggedServiceIds('monolog.processor') as $id => $tags) {
             foreach ($tags as $tag) {
                 if (!empty($tag['channel']) && !empty($tag['handler'])) {
-                    throw new \InvalidArgumentException(sprintf('you cannot specify both the "handler" and "channel" attributes for the "monolog.processor" tag on service "%s"', $id));
+                    throw new \InvalidArgumentException(\sprintf('you cannot specify both the "handler" and "channel" attributes for the "monolog.processor" tag on service "%s"', $id));
```
OskarStark, line 37: Not sure about native function invocation

derrabus: > Not sure about native function invocation 

Can you elaborate? This change looks correct to me.

.github/workflows/ci.yaml
```diff
@@ -5,6 +5,24 @@ on:
...
+        name: Coding Standards
+        steps:
+            -   name: Checkout
+                uses: actions/checkout@v4
+
+            -   name: Setup PHP
+                uses: shivammathur/setup-php@v2
+
+                with:
+                    php-version: 8.1
```
derrabus, line 19: Why not 8.3?

derrabus (approved):

DependencyInjection/Compiler/AddProcessorsPass.php
OskarStark: Sorry, yes it looks correct, but I thought we weren't using it at symfony/symfony. I double checked and we do 👍

derrabus: Yes, we use it for compiler-optimized functions.

.github/workflows/ci.yaml
HypeMC: I just copied the highest PHP version, didn't notice 8.2 & 8.3 are missing, fixed.

derrabus:
Thank you @HypeMC.

PR #489, opened 2024-08-09
Add Sentry `BreadcrumbHandler` support

HypeMC:
Adds support for https://github.com/getsentry/sentry-php/pull/1199.

```yaml
monolog:
  handlers:
    sentry:
      type: sentry
      hub_id: Sentry\State\HubInterface
      level: !php/const Monolog\Logger::ERROR
    sentry_breadcrumb:
      type: sentry_breadcrumb
      sentry_handler: sentry
      level: !php/const Monolog\Logger::INFO

# or if you don't have Sentry bundle installed

monolog:
  handlers:
    sentry:
      type: sentry
      dsn: "https://public@sentry.example.com/1"
      level: !php/const Monolog\Logger::ERROR
    sentry_breadcrumb:
      type: sentry_breadcrumb
      sentry_handler: sentry
      level: !php/const Monolog\Logger::INFO
```

Initially, I considered using the existing `handler` option instead of introducing the new `sentry_handler` option. However, I decided against it because the PSR logger doesn't get auto-enabled when the `handler` option is used:

https://github.com/symfony/monolog-bundle/blob/ed0e4a287282971d110fd8c99d9ac391559a43ce/DependencyInjection/MonologExtension.php#L179-L183

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -236,6 +236,9 @@ use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
  *   - hub_id: Sentry hub custom service id (optional)
  *   - [fill_extra_context]: bool, defaults to false
  *
+ * - sentry_breadcrumb:
+ *   - sentry_handler: the sentry handler's name
+ *
  * - newrelic:
  *   - [level]: level name or int value, defaults to DEBUG
  *   - [bubble]: bool, defaults to true
@@ -584,6 +587,7 @@ class Configuration implements ConfigurationInterface
                 ->booleanNode('persistent')->end() // socket_handler
                 ->scalarNode('dsn')->end() // raven_handler, sentry_handler
                 ->scalarNode('hub_id')->defaultNull()->end() // sentry_handler
+                ->scalarNode('sentry_handler')->defaultNull()->end() // sentry_breadcrumb
                 ->scalarNode('client_id')->defaultNull()->end() // raven_handler, sentry_handler
                 ->scalarNode('auto_log_stacks')->defaultFalse()->end() // raven_handler
                 ->scalarNode('release')->defaultNull()->end() // raven_handler, sentry_handler
@@ -724,6 +728,10 @@ class Configuration implements ConfigurationInterface
             ->validate()
                 ->ifTrue(function ($v) { return 'sentry' === $v['type'] && null !== $v['hub_id'] && null !== $v['client_id']; })
                 ->thenInvalid('You can not use both a hub_id and a client_id in a Sentry handler')
+            ->end()
+            ->validate()
+                ->ifTrue(function ($v) { return 'sentry_breadcrumb' === $v['type'] && !$v['sentry_handler']; })
+                ->thenInvalid('The sentry_handler has to be specified to use a Sentry BreadcrumbHandler')
             ->end()
             ->validate()
                 ->ifTrue(function ($v) { return 'hipchat' === $v['type'] && (empty($v['token']) || empty($v['room'])); })
--- a/DependencyInjection/MonologExtension.php
+++ b/DependencyInjection/MonologExtension.php
@@ -25,6 +25,7 @@ use Symfony\Bridge\Monolog\Processor\SwitchUserTokenProcessor;
 use Symfony\Bridge\Monolog\Processor\TokenProcessor;
 use Symfony\Bridge\Monolog\Processor\WebProcessor;
 use Symfony\Component\Config\FileLocator;
+use Symfony\Component\DependencyInjection\Argument\AbstractArgument;
 use Symfony\Component\DependencyInjection\Argument\BoundArgument;
 use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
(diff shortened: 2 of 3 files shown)


DependencyInjection/MonologExtension.php
```diff
@@ -755,24 +755,35 @@ private function buildHandler(ContainerBuilder $container, $name, array $handler
...
                     }
 
-                    $hub = new Definition(
+                    $hubId = \sprintf('monolog.handler.%s.hub', $name);
+                    $hub = $container->setDefinition($hubId, new Definition(
                         'Sentry\\State\\Hub',
                         [new Reference($clientId)]
-                    );
-                    $container->setDefinition(\sprintf('monolog.handler.%s.hub', $name), $hub);
+                    ));
```
HypeMC, line 762: This never worked correctly. The `monolog.handler.%s.hub` service was registered but never actually used. Instead, the entire definition was passed as an argument, resulting in the creation of two separate services: the unused `monolog.handler.%s.hub` service and an anonymous service that was passed to the handler.

DependencyInjection/Configuration.php
```diff
@@ -725,6 +729,10 @@ public function getConfigTreeBuilder(): TreeBuilder
                 ->ifTrue(function ($v) { return 'sentry' === $v['type'] && null !== $v['hub_id'] && null !== $v['client_id']; })
                 ->thenInvalid('You can not use both a hub_id and a client_id in a Sentry handler')
             ->end()
+            ->validate()
+                ->ifTrue(function ($v) { return 'sentry_breadcrumb' === $v['type'] && empty($v['sentry_handler']); })
```
GromNaN: I think `$v['sentry_handler']` is always defined, and `null` by default.
```suggestion
                ->ifTrue(function ($v) { return 'sentry_breadcrumb' === $v['type'] && !$v['sentry_handler']; })
```

HypeMC: Yes, you're right, fixed.

GromNaN:
> Initially, I considered using the existing handler option instead of introducing the new `sentry_handler` option. However, I decided against it because the PSR logger doesn't get auto-enabled when the handler option is used

Should we make an exception for `sentry_breadcrumb` and always enable the the `process_psr_3_messages` handler?
Using both `handler` and `sentry_handler` with `type: sentry_breadcrumb` would not make sense, does it?

DependencyInjection/Configuration.php
```diff
@@ -594,6 +587,7 @@ public function getConfigTreeBuilder(): TreeBuilder
                 ->booleanNode('persistent')->end() // socket_handler
                 ->scalarNode('dsn')->end() // raven_handler, sentry_handler
                 ->scalarNode('hub_id')->defaultNull()->end() // sentry_handler
+                ->scalarNode('sentry_handler')->defaultNull()->end() // sentry_breadcrumb
```
stof, line 590: We already have a `handler` setting used by all handlers that are wrapping another handler to configure which one it is. It should be reused inside of adding a new setting.

stof:
I don't think we should add support for third-party Monolog handlers in MonologBundle itself, as this will become unmaintainable:

- how do we decide which one we implement ?
- how do we deal with potential BC breaks in those packages for which we have no composer requirements ?

We already support `type: service` to allow defining a service yourselves and then using it as handler (potentially injected into other handlers), which is the way to support community handlers

GromNaN:
The integration with the 3rd party library was added by https://github.com/symfony/monolog-bundle/pull/313. Either we decide to deprecate it in 3.x and remove it in 4.x, or we manage to integrate the features. I prefer the first solution for the rasons you exposed, but we need an upgrade path for users.

Edit: The documentation of the SentryBundle use the `type: service` configuration.
https://docs.sentry.io/platforms/php/guides/symfony/logs/

HypeMC:
I agree that it makes sense not to support 3rd-party handlers. Sentry's is currently the only one, so it's an oddball.

However, I don't think 3rd-party handlers are the only issue. For example, this handler seems to be broken:

```php
case 'rollbar':
    if (!empty($handler['id'])) {
        $rollbarId = $handler['id'];
    } else {
        $config = $handler['config'] ?: [];
        $config['access_token'] = $handler['token'];
        $rollbar = new Definition('RollbarNotifier', [
            $config,
        ]);
        $rollbarId = 'monolog.rollbar.notifier.'.sha1(json_encode($config));
        $rollbar->setPublic(false);
        $container->setDefinition($rollbarId, $rollbar);
    }

    $definition->setArguments([
        new Reference($rollbarId),
        $handler['level'],
        $handler['bubble'],
    ]);
    break;
```

The `RollbarNotifier` class doesn't exist in the current version of the `rollbar/rollbar` library, and Monolog v3's `RollbarHandler` doesn't support it either. I think referencing any 3rd-party classes in the code could lead to potential BC breaks, so maybe all of this should be removed?

stof:
@HypeMC this is because this code is related to support of the Monolog RollbarHandler of older versions of Monolog (which we'll be able to clean in 4.0).
Monolog itself has a policy to not accept new handlers in core that require third-party dependencies. But I think that we should keep support for those handlers in MonologBundle as long as they are part of Monolog itself.

HypeMC:
@stof Yes, we should keep support for the handlers themselves. I was thinking we could maybe drop support for registering the 3rd-party classes and just keep the option of providing the service ID (in other words, drop the `else` part from my previous example).

HypeMC:
Closing this one as we don't want to support third party handlers.

PR #490, opened 2024-10-15
Fix `file_permission` type in the Configuration

php4fan:
Fixed wrong parameter type in documentation of rotating_file handler


DependencyInjection/Configuration.php
```diff
@@ -63,7 +63,7 @@
  *   - [max_files]: files to keep, defaults to zero (infinite)
  *   - [level]: level name or int value, defaults to DEBUG
  *   - [bubble]: bool, defaults to true
- *   - [file_permission]: string|null, defaults to null
+ *   - [file_permission]: int|null, defaults to null (0644)
```
stof, line 66:
```suggestion
 *   - [file_permission]: int|null, defaults to null (0o644)
```

We need to use the proper octal notation if using octal for the default value

GromNaN (approved):

GromNaN:
Thank you @php4fan.

PR #492, opened 2024-11-03
Add support for configuring Gelf encoders in Monolog configuration

Tib-z:
This commit introduces the ability to specify custom encoders (`json` and `compressed_json`) for Gelf publishers in Monolog's configuration.

Overview
This PR introduces an enhancement to the Symfony Monolog Bundle by adding a configuration option for setting the message encoder for the GELF handler's publisher. This change addresses a backward compatibility issue introduced in the 2.0 update of the graylog2/gelf-php package, which shifted the default message encoder from CompressedJsonEncoder to JsonEncoder. Thus an upgrade silently breakes the logging.

https://github.com/bzikarsky/gelf-php/commit/f2625cf008289e9a5004bf55eaf6e95dcbcbc6c0#diff-9520be6ea98f5ded11cbe62c500f1a8fef819810d8836d2b15189cdbc073900cR32
https://github.com/bzikarsky/gelf-php/commit/f2625cf008289e9a5004bf55eaf6e95dcbcbc6c0#diff-72a9d13c23b7d615833219f0dfc92be317d9a4981de7b4747e287db62aaa1079L69

In a Symfony project, to override this default behavior, users had to manually construct and define the transport, publisher, and encoder as a service and then use it in the Monolog configuration. This approach, while functional, is cumbersome. It would be more user-friendly to enable setting the encoder via a simple configuration option. The proposed change would be:

```
            gelf:
                type: gelf
                level: debug
                publisher:
                    hostname: '%env(resolve:LOGSTASH_HOSTNAME)%'
                    port: '%env(resolve:LOGSTASH_PORT)%'
                    encoder: compressed_json  # Optional: 'json' or 'compressed_json'
```
This new encoder configuration is optional and does not have an explicit default value, thereby preserving the existing behavior (albeit the recent default may not align with prior expectations).

While it might be tempting to expose more variability (like using class names or service IDs directly), overly customizable options may not add practical value and could complicate the configuration unnecessarily, 'coz YAGNI :)

### DIFF
--- a/src/DependencyInjection/MonologExtension.php
+++ b/src/DependencyInjection/MonologExtension.php
@@ -23,6 +23,7 @@ use Symfony\Bridge\Monolog\Handler\FingersCrossed\HttpCodeActivationStrategy;
 use Symfony\Bridge\Monolog\Processor\SwitchUserTokenProcessor;
 use Symfony\Bridge\Monolog\Processor\TokenProcessor;
 use Symfony\Bridge\Monolog\Processor\WebProcessor;
+use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\Argument\BoundArgument;
 use Symfony\Component\DependencyInjection\ChildDefinition;
@@ -221,10 +222,28 @@ class MonologExtension extends Extension
                     ]);
                     $transport->setPublic(false);
 
+                    if (isset($handler['publisher']['encoder'])) {
+                        if ('compressed_json' === $handler['publisher']['encoder']) {
+                            $encoderClass = 'Gelf\Encoder\CompressedJsonEncoder';
+                        } elseif ('json' === $handler['publisher']['encoder']) {
+                            $encoderClass = 'Gelf\Encoder\JsonEncoder';
+                        } else {
+                            throw new InvalidConfigurationException('The gelf message encoder must be either "compressed_json" or "json".');
+                        }
+
+                        $encoder = new Definition($encoderClass);
+                        $encoder->setPublic(false);
+
+                        $transport->addMethodCall('setMessageEncoder', [$encoder]);
+                    }
+
                     $publisher = new Definition('Gelf\Publisher', []);
                     $publisher->addMethodCall('addTransport', [$transport]);
                     $publisher->setPublic(false);
                 } elseif (class_exists('Gelf\MessagePublisher')) {
+                    if (isset($handler['publisher']['encoder']) && 'compressed_json' !== $handler['publisher']['encoder']) {
+                        throw new InvalidConfigurationException('The Gelf\MessagePublisher publisher supports only the compressed json encoding. Omit the option to use the default encoding or use "compressed_json" as the encoder option.');
+                    }
                     $publisher = new Definition('Gelf\MessagePublisher', [
                         $handler['publisher']['hostname'],
                         $handler['publisher']['port'],
--- a/src/DependencyInjection/MonologExtension.php
+++ b/src/DependencyInjection/MonologExtension.php
@@ -228,7 +228,7 @@ class MonologExtension extends Extension
                         } elseif ('json' === $handler['publisher']['encoder']) {
                             $encoderClass = 'Gelf\Encoder\JsonEncoder';
                         } else {
-                            throw new InvalidConfigurationException('The gelf message encoder must be either "compressed_json" or "json".');
+                            throw new \RuntimeException('The gelf message encoder must be either "compressed_json" or "json".');
                         }
 
                         $encoder = new Definition($encoderClass);
--- a/src/DependencyInjection/MonologExtension.php
+++ b/src/DependencyInjection/MonologExtension.php
@@ -242,7 +242,7 @@ class MonologExtension extends Extension
                     $publisher->setPublic(false);
                 } elseif (class_exists('Gelf\MessagePublisher')) {
                     if (isset($handler['publisher']['encoder']) && 'compressed_json' !== $handler['publisher']['encoder']) {
-                        throw new InvalidConfigurationException('The Gelf\MessagePublisher publisher supports only the compressed json encoding. Omit the option to use the default encoding or use "compressed_json" as the encoder option.');
+                        throw new \RuntimeException('The Gelf\MessagePublisher publisher supports only the compressed json encoding. Omit the option to use the default encoding or use "compressed_json" as the encoder option.');
                     }
                     $publisher = new Definition('Gelf\MessagePublisher', [
                         $handler['publisher']['hostname'],
--- a/src/DependencyInjection/MonologExtension.php
+++ b/src/DependencyInjection/MonologExtension.php
@@ -23,7 +23,6 @@ use Symfony\Bridge\Monolog\Handler\FingersCrossed\HttpCodeActivationStrategy;
 use Symfony\Bridge\Monolog\Processor\SwitchUserTokenProcessor;
 use Symfony\Bridge\Monolog\Processor\TokenProcessor;
 use Symfony\Bridge\Monolog\Processor\WebProcessor;
-use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\Argument\BoundArgument;
 use Symfony\Component\DependencyInjection\ChildDefinition;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Drop support for Symfony < 6.4
 * Add TelegramBotHandler `topic` support
 * Deprecate `sentry` and `raven` handler, use a `service` handler with [`sentry/sentry-symfony`](https://docs.sentry.io/platforms/php/guides/symfony/logs/) instead
+* Add configuration for Gelf encoders
 
 ## 3.10.0 (2023-11-06)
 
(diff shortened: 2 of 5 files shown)


Tib-z:
Well class_alias(\stdClass::class, 'Gelf\Transport\UdpTransport'); in the tests obviously does not work in <8.3 😓 
Generally, workarounds involving class existence checks are hacky, but I wanted to ensure the tests provided meaningful coverage.

To address the compatibility issue, I've added a simple dummy class as a workaround. Another option could be creating the class dynamically with eval, though that also smells... A potential refactor might involve wrapping class_exists checks and mocking them, but I'm unsure if the benefit would justify the effort.

Let me know your thoughts

DependencyInjection/Configuration.php
```diff
@@ -50,7 +50,12 @@
  *   - [bubble]: bool, defaults to true
  *
  * - gelf:
- *   - publisher: {id: ...} or {hostname: ..., port: ..., chunk_size: ...}
+ *   - publiser:
```
derrabus, line 53:
```suggestion
 *   - publisher:
```

DependencyInjection/Configuration.php
```diff
@@ -50,7 +50,12 @@
...
+ *      - id: string, service id of a publisher implementation, optional if hostname is given
+ *      - hostname: string, optional if id is given
+ *      - [port]: int, defaults to 12201
+ *      - [chunk_size]: int, defaults to 1420
+ *      - [encoder]: string, its value can be 'json' or 'compressed_json'
```
derrabus, line 58: Before your change, it was clear that I should either provide a service ID or configure the connection by specifying the hostname, port etc. The host name is not "optional if id is given" as you wrote, it's ignored entirely if a service ID has been configured.

derrabus (commented):
Thank you for your PR. Can you elaborate a bit more why I should care about which encoder is used? If the protocol allows both and thus a backend implementing the protocol should support both, I don't understand why it makes a difference for my app.

I also wonder how common your problem is. If this only affects a handful of projects, we could recommend to configure the publisher service manually and configure a service ID instead of connection details.

Tib-z:
In our infra, in the logstash pipeline comporessed json is a contraint, does not work without comporession, so the change in the gelf package broke the logging. Likely it's a specific setup, not sure exactly how is configured, I'll try to figure out (it's not maintained by me/my team) and see if it could affect others as well.

As the encoder is a dependency and straightforward to change it, it was logical move to expose it as a monolog config option and make the implementation easier. In a microservice architecture there are 30-40 services where have to write boilerplate service definitions vs add one option, however I got your point, the service id workaround is still an option.

DependencyInjection/Configuration.php
Tib-z: Typo fixed 😄

derrabus:
That makes perfect sense. Thank you for the explanation.

DependencyInjection/Configuration.php
Tib-z:
```
 * - publisher: (one of the following configurations)
 *      # Option 1: Service-based configuration
 *      - id: string, service id of a publisher implementation
 *      
 *      # Option 2: Direct connection configuration
 *      - hostname: string, server hostname
 *      - [port]: int, server port (default: 12201)
 *      - [chunk_size]: int, UDP packet size (default: 1420)
 *      - [encoder]: string, encoding format ('json' or 'compressed_json')
```

WDYT about this approach?

Tests/DependencyInjection/MonologExtensionTest.php
```diff
@@ -186,6 +187,97 @@ public function testExceptionWhenUsingGelfWithoutPublisherHostname()
         $loader->load([['handlers' => ['gelf' => ['type' => 'gelf', 'publisher' => []]]]], $container);
     }
 
+    public function testExceptionWhenUsingLegacyGelfImplementationWithUnsupportedEncoder(): void
+    {
+        if (!class_exists('Gelf\MessagePublisher')) {
+            class_alias(DummyClassForClassExistsCheck::class, 'Gelf\MessagePublisher');
```
stof, line 193: Doing such class aliases is a hack, and won't work properly. If the `testExceptionWhenUsingGelfWithInvalidEncoder` test is run first, this test will never reach the code path guarded by `class_exists('Gelf\MessagePublisher')` (because the actual guard to reach that code path is `!class_exists('Gelf\Transport\UdpTransport') && class_exists('Gelf\MessagePublisher')` and you cannot remove the other alias).

Tests should be written against the installed dependencies, without such hacks, and skipped in case they are not relevant.

Tib-z: Thanks for the feedback; I completely agree. While this approach is hackish, it did feel wrong that the changes couldn’t be properly tested. I initially used these as part of my testing process and I left them in the PR...

As I mentioned in my previous comment:

> Generally, workarounds involving class existence checks are hacky, but I wanted to ensure the tests provided meaningful coverage.

I'll remove them.

DependencyInjection/Configuration.php
Tib-z: @derrabus  So, I went this way. Any other suggestion for this PR?

DependencyInjection/Configuration.php
```diff
@@ -50,9 +50,17 @@
...
+ *   # Option 1: Service-based configuration
+ *   - id: string, service id of a publisher implementation
+ *
+ *   # Option 2: Direct connection configuration
+ *   - hostname: string, server hostname
+ *   - [port]: int, server port (default: 12201)
+ *   - [chunk_size]: int, UDP packet size (default: 1420)
+ *   - [encoder]: string, encoding format ('json' or 'compressed_json')
+ * - [level]: level name or int value, defaults to DEBUG
+ * - [bubble]: bool, defaults to true
```
HypeMC: The indentation is wrong:

```suggestion
 *   - publisher: (one of the following configurations)
 *     # Option 1: Service-based configuration
 *     - id: string, service id of a publisher implementation
 *
 *     # Option 2: Direct connection configuration
 *     - hostname: string, server hostname
 *     - [port]: int, server port (default: 12201)
 *     - [chunk_size]: int, UDP packet size (default: 1420)
 *     - [encoder]: string, encoding format ('json' or 'compressed_json')
 *   - [level]: level name or int value, defaults to DEBUG
 *   - [bubble]: bool, defaults to true
```

DependencyInjection/Configuration.php
```diff
@@ -816,6 +824,7 @@ private function addGelfSection(ArrayNodeDefinition $handerNode)
                         ->scalarNode('hostname')->end()
                         ->scalarNode('port')->defaultValue(12201)->end()
                         ->scalarNode('chunk_size')->defaultValue(1420)->end()
+                        ->scalarNode('encoder')->end()
```
HypeMC: Perhaps an `enumNode` would be better here:

```suggestion
                        ->enumNode('encoder')->values(['json', 'compressed_json'])->end()
```

DependencyInjection/MonologExtension.php
```diff
@@ -227,10 +228,28 @@ private function buildHandler(ContainerBuilder $container, $name, array $handler
                     ]);
                     $transport->setPublic(false);
 
+                    if (isset($handler['publisher']['encoder'])) {
+                        if ('compressed_json' === $handler['publisher']['encoder']) {
+                            $encoderClass = 'Gelf\Encoder\CompressedJsonEncoder';
+                        } elseif ('json' === $handler['publisher']['encoder']) {
+                            $encoderClass = 'Gelf\Encoder\JsonEncoder';
+                        } else {
+                            throw new InvalidConfigurationException('The gelf message encoder must be either "compressed_json" or "json".');
```
HypeMC: An `InvalidConfigurationException` is usually thrown by the configuration, not the extension:

```suggestion
                            throw new \RuntimeException('The gelf message encoder must be either "compressed_json" or "json".');
```

HypeMC (commented):
Thanks, I've tested the changes and everything seems to work as expected. Just have a few minor comments.

Tib-z:
> Thanks, I've tested the changes and everything seems to work as expected. Just have a few minor comments.

Thanks for the review.

DependencyInjection/MonologExtension.php
```diff
@@ -227,10 +228,28 @@ private function buildHandler(ContainerBuilder $container, $name, array $handler
...
+
+                        $transport->addMethodCall('setMessageEncoder', [$encoder]);
+                    }
+
                     $publisher = new Definition('Gelf\Publisher', []);
                     $publisher->addMethodCall('addTransport', [$transport]);
                     $publisher->setPublic(false);
                 } elseif (class_exists('Gelf\MessagePublisher')) {
+                    if (isset($handler['publisher']['encoder']) && 'compressed_json' !== $handler['publisher']['encoder']) {
+                        throw new InvalidConfigurationException('The Gelf\MessagePublisher publisher supports only the compressed json encoding. Omit the option to use the default encoding or use "compressed_json" as the encoder option.');
```
HypeMC:
```suggestion
                        throw new \RuntimeException('The Gelf\MessagePublisher publisher supports only the compressed json encoding. Omit the option to use the default encoding or use "compressed_json" as the encoder option.');
```

HypeMC (commented):
Hi @Tib-z, could you please rebase onto the `3.x` branch? We've updated the directory structure.

src/DependencyInjection/MonologExtension.php
```diff
@@ -23,6 +23,7 @@
 use Symfony\Bridge\Monolog\Processor\SwitchUserTokenProcessor;
 use Symfony\Bridge\Monolog\Processor\TokenProcessor;
 use Symfony\Bridge\Monolog\Processor\WebProcessor;
+use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
```
HypeMC: This is now unused.

HypeMC (approved):

GromNaN (approved):

GromNaN:
Thank you @Tib-z.

PR #493, opened 2024-11-03
Fix compatibility with PHPUnit 11

derrabus:
This PR fixes a couple of compatibility issues with PHPUnit 11:

* Data provider methods must be static.
* The mock builder does not have a `setMethods()` method anymore.
* Abstract test classes must not use the suffix `Test`.

Note: The tests are still run with PHPUnit 8.5 and 9.6 though. The PR merely enables us to use PHPUnit 11 in the future.

### DIFF
--- a/Tests/DependencyInjection/Compiler/AddSwiftMailerTransportPassTest.php
+++ b/Tests/DependencyInjection/Compiler/AddSwiftMailerTransportPassTest.php
@@ -28,10 +28,7 @@ class AddSwiftMailerTransportPassTest extends TestCase
 
     private $definition;
 
-    /**
-     * @before
-     */
-    protected function doSetUp()
+    protected function setUp(): void
     {
         $this->compilerPass = new AddSwiftMailerTransportPass();
         $this->definition = new Definition(null, [new Reference('swiftmailer')]);
--- a/Tests/DependencyInjection/ConfigurationTest.php
+++ b/Tests/DependencyInjection/ConfigurationTest.php
@@ -42,7 +42,7 @@ class ConfigurationTest extends TestCase
         $this->assertFalse($config['handlers']['foobar']['nested']);
     }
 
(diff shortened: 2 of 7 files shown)


Tests/DependencyInjection/Compiler/AddSwiftMailerTransportPassTest.php
```diff
@@ -26,19 +28,16 @@ class AddSwiftMailerTransportPassTest extends TestCase
...
         $this->compilerPass = new AddSwiftMailerTransportPass();
-        $this->definition = $this->getMockBuilder('\Symfony\Component\DependencyInjection\Definition')->getMock();
+        $this->definition = $this->getMockBuilder(Definition::class)->getMock();
         $this->definition->expects($this->any())
             ->method('getArgument')
             ->with(0)
             ->willReturn(new Reference('swiftmailer'));
-        $this->container = $this->getMockBuilder('\Symfony\Component\DependencyInjection\ContainerBuilder')
-            ->setMethods(['getParameter', 'getDefinition', 'hasDefinition', 'addMethodCall'])->getMock();
+        $this->container = $this->getMockBuilder(ContainerBuilder::class)
```
xabbuh, line 39: Do we really have to mock the `ContainerBuilder` class? What about using a "real" instance with "real" definitions instead?

derrabus: We probably don't. But I did not want to rewrite the whole test, tbh. 😓 

But I can look into it.

xabbuh: we can do that first (see #496) and then you could rebase here

xabbuh (approved):

xabbuh:
Thank you Alexander.

PR #494, opened 2024-11-03
Drop support for Monolog < 3.5

derrabus:
Monolog 1 is more or less unmaintained these days. Dropping support for that old branch allows us to clean up the bundle a bit and reduce the size of our CI matrix. Projects that really want to keep using Monolog 1 can still install an old version of the bundle.

edit: I've also removed code paths exclusive to Monolog 2, given that version 3 is the minimum now.


lyrixx (approved):
Thanks 👍🏼

GromNaN (approved):

GromNaN:
Hi @derrabus! Would you like to rebase this PR on 4.x?

derrabus:
Done.

derrabus:
I've also removed code paths exclusive to Monolog 2, given that version 3 is the minimum now.

src/DependencyInjection/MonologExtension.php
```diff
@@ -527,43 +507,6 @@ private function buildHandler(ContainerBuilder $container, $name, array $handler
...
-            case 'swift_mailer':
-                $mailer = $handler['mailer'] ?: 'mailer';
-                if (isset($handler['email_prototype'])) {
-                    if (!empty($handler['email_prototype']['method'])) {
-                        $prototype = [new Reference($handler['email_prototype']['id']), $handler['email_prototype']['method']];
-                    } else {
-                        $prototype = new Reference($handler['email_prototype']['id']);
-                    }
-                } else {
-                    $messageFactory = new Definition('Symfony\Bundle\MonologBundle\SwiftMailer\MessageFactory');
```
GromNaN, line 539: The class `SwiftMailer\MessageFactory` is no longer used.

derrabus:
We had some logic to support Monolog before 3.5 which I would also consider obsolete now. I've removed that as well.

GromNaN (approved):

GromNaN:
Thank you Alexander.

PR #495, opened 2024-11-04
Fix test with `symfony/config:^7.2`

alexandre-daubois:
Following https://github.com/symfony/symfony/pull/58490#issuecomment-2453496578. Until the Symfony PR, null wasn't supported in boolean nodes and defaulted to true.

### DIFF
--- a/Tests/DependencyInjection/ConfigurationTest.php
+++ b/Tests/DependencyInjection/ConfigurationTest.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Bundle\MonologBundle\Tests\DependencyInjection;
 
+use Composer\InstalledVersions;
 use Monolog\Logger;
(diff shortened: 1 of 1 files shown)


derrabus:
Thank you. Your change makes our test suite green again, which is great of course. But, are we sure this subtle change in behavior won't break stuff downstream?

alexandre-daubois:
I had a look while creating the PR, and it seems the `null` possibility was actually handled, even if not possible until now : https://github.com/symfony/monolog-bundle/blob/ed0e4a287282971d110fd8c99d9ac391559a43ce/DependencyInjection/MonologExtension.php#L180

derrabus (approved):

derrabus:
Thank you @alexandre-daubois.

PR #501, opened 2025-02-04
fix: correct service parameters for `elastic_search` type handlers

### LINKED ISSUE #500: `elastic_search` Handler Does Not Work (Invalid Factory Parameters)
As noted [here](https://github.com/symfony/monolog-bundle/pull/438#issuecomment-1542839322), using [`type: elastic_search`](https://github.com/symfony/monolog-bundle/blob/master/DependencyInjection/Configuration.php#L83) for a handler does not work.  We attempt to [pass the host as `host`](https://github.com/symfony/monolog-bundle/blob/master/DependencyInjection/MonologExtension.php#L294), but the [`ClientBuilder::fromConfig`](https://github.com/elastic/elasticsearch-php/blob/main/src/ClientBuilder.php#L144) from upstream expects a plural [`hosts` array](https://github.com/elastic/elasticsearch-php/blob/main/src/ClientBuilder.php#L213).

This applies to both [v7 and v8](https://github.com/Seldaek/monolog/blob/main/composer.json#L23). There are no workarounds, except to install the Elastica client instead (v7 due to https://github.com/symfony/monolog-bundle/issues/499).

We have no tests for `elastic_search` handlers.

Kenneth-Sills:
Closes #500

### DIFF
--- a/src/DependencyInjection/MonologExtension.php
+++ b/src/DependencyInjection/MonologExtension.php
@@ -304,7 +304,7 @@ class MonologExtension extends Extension
                         $factory = class_exists('Elastic\Elasticsearch\ClientBuilder') ? 'Elastic\Elasticsearch\ClientBuilder' : 'Elasticsearch\ClientBuilder';
                         $client->setFactory([$factory, 'fromConfig']);
                         $clientArguments = [
-                            'host' => $handler['elasticsearch']['host'],
+                            'hosts' => [$handler['elasticsearch']['host']],
                         ];
 
                         if (isset($handler['elasticsearch']['user'], $handler['elasticsearch']['password'])) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Add TelegramBotHandler `topic` support
 * Deprecate `sentry` and `raven` handler, use a `service` handler with [`sentry/sentry-symfony`](https://docs.sentry.io/platforms/php/guides/symfony/logs/) instead
 * Add configuration for Gelf encoders
+* Fix `host` configuration for `elastic_search` handler
 
 ## 3.10.0 (2023-11-06)
 


DependencyInjection/MonologExtension.php
```diff
@@ -292,7 +292,7 @@ private function buildHandler(ContainerBuilder $container, $name, array $handler
                         $factory = class_exists('Elastic\Elasticsearch\ClientBuilder') ? 'Elastic\Elasticsearch\ClientBuilder' : 'Elasticsearch\ClientBuilder';
                         $client->setFactory([$factory, 'fromConfig']);
                         $clientArguments = [
-                            'host' => $handler['elasticsearch']['host'],
+                            'hosts' => [$handler['elasticsearch']['host']],
```
stof, line 295: This is not valid. The arguments need to be different for V7 or v8 (see just above where we already detect version 8+ for other differences)

Kenneth-Sills: Like I said in the ticket, this applies to both v7 and v8. [`ClientBuilder::fromConfig`(https://github.com/elastic/elasticsearch-php/blob/v7.17.2/src/Elasticsearch/ClientBuilder.php#L183) works the same in both versions by proxying to the `set*` method and the only such method that exists is [`setHosts`](https://github.com/elastic/elasticsearch-php/blob/v7.17.2/src/Elasticsearch/ClientBuilder.php#L392), not `setHost`.

In my demo project I just ran `composer require elasticsearch/elasticsearch=^7` (using Elastica, so no v8), then set the monolog config like:

```yaml
  monolog:
    handlers:
      main:
        type: fingers_crossed
        handler: elastic_search
        action_level: error
        excluded_http_codes: [404, 405]
        buffer_size: 50
      elastic_search:
        type: elastic_search
        elasticsearch:
          host: '%elastic_domain%'
          port: '%elastic_port%'
        index: '%elastic_index_monolog%'
```

Finally, I ran `./bin/console cache:clear` (after `rm -rf var/cache/*` for good measure). And it also fails with:

```yaml
In ClientBuilder.php line 217:

  Unknown parameters provided: host  
```

Kenneth-Sills: @stof Bump!

*(Let me know if y'all don't like bumps.)*

src/DependencyInjection/MonologExtension.php
```diff
@@ -292,7 +292,7 @@ private function buildHandler(ContainerBuilder $container, $name, array $handler
                         $factory = class_exists('Elastic\Elasticsearch\ClientBuilder') ? 'Elastic\Elasticsearch\ClientBuilder' : 'Elasticsearch\ClientBuilder';
                         $client->setFactory([$factory, 'fromConfig']);
                         $clientArguments = [
-                            'host' => $handler['elasticsearch']['host'],
+                            'hosts' => [$handler['elasticsearch']['host']],
```
GromNaN: Looking at the method `ClientBuilder::fromConfig` of all versions of [elasticsearch-php](https://github.com/elastic/elasticsearch-php/tree/main), I see that the parameter has always been the plural `hosts`. This update looks good to me.

The PR needs to be rebased on 3.x.

GromNaN (approved):

GromNaN:
Thanks for fixing this bug @Kenneth-Sills.

PR #504, opened 2025-03-17
feat: console.interactive_only flag

### LINKED ISSUE #58715: [Console] allow the console logging handler to work and prevent propagation only when interactive
### Description

Let's assume this setup

```yaml
monolog:
    handlers:
        console:
            type: console
            interactive_only: true # this is new
        logstash:
            type: stream
            path: "php://stderr"
            formatter: 'monolog.formatter.logstash'
```

Now let's assume we're running some command meant for non-interactive use, like `messenger:consume`, see #57776.

When running the command using logging interactively (user running the command and watching the output) currently, you'll get both the command output, console output and the logstash JSON, which is creating a giant mess of output.

What would be better is to allow the logger to know if there's a user watching the output (command is in an interactive shell or not) and then:

1. output `console` and skip `logstash` (prevent bubble) when interactive
2. skip `console` and only output `logstash` when not interactive

This would mean the logging is adjusted to the current runtime better automatically.

### Example

_No response_

dkarlovi:
Note: needs https://github.com/symfony/symfony/pull/59955

Closes https://github.com/symfony/symfony/issues/58715

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Bundle\MonologBundle\DependencyInjection;
 
+use Composer\InstalledVersions;
 use Monolog\Logger;
 use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
@@ -40,6 +41,7 @@ use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
  *   - [level]: level name or int value, defaults to DEBUG
  *   - [bubble]: bool, defaults to true
  *   - [console_formatter_options]: array
+ *   - [interactive_only]: bool, defaults to false
  *
  * - firephp:
  *   - [level]: level name or int value, defaults to DEBUG
@@ -448,6 +450,7 @@ class Configuration implements ConfigurationInterface
                 ->scalarNode('priority')->defaultValue(0)->end()
                 ->scalarNode('level')->defaultValue('DEBUG')->end()
                 ->booleanNode('bubble')->defaultTrue()->end()
+                ->booleanNode('interactive_only')->defaultFalse()->end()
                 ->scalarNode('app_name')->defaultNull()->end()
                 ->booleanNode('fill_extra_context')->defaultFalse()->end() // sentry
                 ->booleanNode('include_stacktraces')->defaultFalse()->end()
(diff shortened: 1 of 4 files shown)


DependencyInjection/MonologExtension.php
```diff
@@ -203,6 +203,8 @@ private function buildHandler(ContainerBuilder $container, $name, array $handler
                     $handler['bubble'],
                     $handler['verbosity_levels'] ?? [],
                     $handler['console_formatter_options'],
+                    null,
+                    $handler['interactive_only'],
```
dkarlovi, line 207: This will be false by default and they will get added even if your Symfony Monolog bridge is too old, but it doesn't hurt, they'll just get ignored.

DependencyInjection/Configuration.php
```diff
@@ -796,6 +797,10 @@ public function getConfigTreeBuilder(): TreeBuilder
                 ->ifTrue(function ($v) { return 'server_log' === $v['type'] && empty($v['host']); })
                 ->thenInvalid('The host has to be specified to use a ServerLogHandler')
             ->end()
+            ->validate()
+                ->ifTrue(function ($v) { return 'console' !== $v['type'] && $v['interactive_only']; })
```
dkarlovi, line 801: @stof is this where we check for bridge version being >= 7.3, is there an example in the bundle already?

dkarlovi: @stof do you have any guidance here?

DependencyInjection/Configuration.php
```diff
@@ -796,6 +797,10 @@ public function getConfigTreeBuilder(): TreeBuilder
...
+                ->thenInvalid('The interactive_only flag is only available with the console handler')
```
stof, line 802: We generally don't throw errors for flags that are ignored for other types, as this makes it a lot harder to override the type of handler used for another environment (we would have to explicitly reset other fields).

All our validation rules are about settings that are required for a given type.

DependencyInjection/Configuration.php
stof: either find a class or interface allowing to easily detect that with a `class_exists` or `interface_exists` check.

Otherwise, an alternative is to add a dependency on `composer-runtime-api` and use their `InstalledVersions` API

DependencyInjection/Configuration.php
dkarlovi: That means you can set any setting for any type though, but OK I guess, I'll remove it.

dkarlovi:
@GromNaN could you also check this one since it's related to the other one?

DependencyInjection/Configuration.php
```diff
@@ -796,6 +799,10 @@ public function getConfigTreeBuilder(): TreeBuilder
                 ->ifTrue(function ($v) { return 'server_log' === $v['type'] && empty($v['host']); })
                 ->thenInvalid('The host has to be specified to use a ServerLogHandler')
             ->end()
+            ->validate()
+                ->ifTrue(function ($v) { return $v['interactive_only'] && version_compare(InstalledVersions::getVersion('symfony/monolog-bridge'), '7.3.0', '<'); })
+                ->thenInvalid('The interactive_only flag is available with symfony/monolog-bridge 7.3.0 or higher')
```
GromNaN:
```suggestion
                ->thenInvalid('The interactive_only flag requires symfony/monolog-bridge 7.4 or higher')
```

GromNaN (commented):
SGTM. Will merge after https://github.com/symfony/symfony/pull/59955

GromNaN:
Thank you @dkarlovi.

PR #509, opened 2025-08-25
Convert config from XML to PHP

nicolas-grekas:
PHP is nicer than XML and https://github.com/symfony/symfony/pull/60568 might make this required at some point.


Resources/config/monolog.php
```diff
@@ -0,0 +1,66 @@
...
+use Psr\Log\LoggerInterface;
+use Symfony\Bridge\Monolog\Handler\FingersCrossed\NotFoundActivationStrategy;
+use Symfony\Component\HttpClient\HttpClient;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+return static function (ContainerConfigurator $container) {
+    $container->services()
+
+        ->alias('logger', 'monolog.logger')
+        ->alias(LoggerInterface::class, 'logger')
```
GromNaN, line 35: `logger` was a public and `LoggerInterface::class` not public. Do you need to make one public or the other private?

derrabus: Sure that `logger` was public? There was no explicit visibility configured and the default should be "not public".

PR #513, opened 2025-08-30
Don't add processor to all loggers if tags specify a channel or handler

HypeMC:
Similar to https://github.com/symfony/symfony/pull/61526.

When you add a custom processor to a specific channel or handler, it gets added to **all** channels because `ProcessorInterface` is [registered for autoconfiguration](https://github.com/symfony/monolog-bundle/blob/58b98dcf5d3c818679f078d514284b16cc986d34/DependencyInjection/MonologExtension.php#L106-L108).

Here's a simple example:

```php
#[AsMonologProcessor('my_channel')]
final class MyProcessor implements ProcessorInterface
{
    public function __invoke(LogRecord $record) {}
}
```

You'd expect this processor to be registered only for `my_channel`, but due to autoconfiguration it's added to every channel.

The workaround is to disable autoconfiguration and add an explicit tag.

```yaml
services:
    App\MyProcessor:
        autoconfigure: false
        tags:
            - monolog.processor: { channel: 'my_channel' }
```

This PR ensures a processor is not added to all channels when an explicit channel or handler is configured.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Migrate services configuration to PHP
 * Add `console.interactive_only` flag
 * Add `slack.exclude_fields` and `slackwebhook.exclude_fields` configuration
+* Add a processor to all loggers only when tags do not specify a channel or handler
 
 ## 3.10.0 (2023-11-06)
 
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,8 @@
         "symfony/config": "^5.4 || ^6.0 || ^7.0",
         "symfony/dependency-injection": "^5.4 || ^6.0 || ^7.0",
         "symfony/http-kernel": "^5.4 || ^6.0 || ^7.0",
-        "symfony/monolog-bridge": "^5.4 || ^6.0 || ^7.0"
+        "symfony/monolog-bridge": "^5.4 || ^6.0 || ^7.0",
+        "symfony/polyfill-php84": "^1.30"
     },
     "require-dev": {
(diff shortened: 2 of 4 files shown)


HypeMC:
~~CI failure is false positive, trailing_comma_in_multiline is not supported in PHP 7.2.~~

Decided to put in in one line instead.

Tests/DependencyInjection/Compiler/AddProcessorsPassTest.php
```diff
@@ -65,6 +68,61 @@ public function testFailureOnHandlerWithoutPushProcessor()
...
+    /**
+     * @dataProvider provideEmptyTagsData
+     */
+    public function testEmptyTagsAreIgnoredWhenNonEmptyArePresent(
+        array $tagAttributesList,
+        array $expectedLoggerCalls,
+        array $expectedMyChannelLoggerCalls
+    ) {
+        $container = new ContainerBuilder();
+        $loader = new PhpFileLoader($container, new FileLocator(__DIR__.'/../../../Resources/config'));
```
GromNaN:
```suggestion
        $loader = new PhpFileLoader($container, new FileLocator(__DIR__.'/../../../../config'));
```

GromNaN (approved):
LGTM.
@HypeMC can you rebase the PR on 3.x?

Tests/DependencyInjection/Compiler/AddProcessorsPassTest.php
HypeMC: Done

GromNaN:
Thank you @HypeMC.

PR #518, opened 2025-09-24
Run CI with various versions of Symfony

GromNaN:
| Q             | A
| ------------- | ---
| Branch?       | 3.x
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

Ensure the bundle is compatible with all the supported Symfony versions.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,21 +28,33 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php: [ '7.2', '7.3', '7.4', '8.0', '8.1', '8.2', '8.3' ]
-                monolog: [ '1.*', '2.*' ]
+                php: [ '8.1', '8.2', '8.3', '8.4' ]
+                monolog: [ '2.*' ]
+                symfony: [ '6.4.*' ]
                 include:
+                    -   php: '7.2'
+                        symfony: '5.4.*'
+                        monolog: '1.*'
                     -   php: '7.4'
                         deps: lowest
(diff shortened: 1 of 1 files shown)


HypeMC (approved):

.github/workflows/ci.yaml
```diff
@@ -28,21 +28,33 @@ jobs:
...
                         deprecations: max[self]=0
                     -   php: '8.1'
                         monolog: '3.*'
+                        symfony: '6.4.*'
                     -   php: '8.2'
                         monolog: '3.*'
+                        symfony: '7.3.*'
                     -   php: '8.3'
                         monolog: '3.*'
+                        symfony: '7.3.*'
```
stof, line 50: This is the wrong approach to solve this problem IMO, as it means that we have to remember to update the CI of monolog-bundle every 6 months when Symfony releases new versions, otherwise we don't test latest versions (while our requirements allow them).

A cleaner solution is to have most job run with original composer requirements (so no `SYMFONY_REQUIRE`), and to add extra jobs forcing a Symfony version for the supported LTS versions (which need maintenance only every 2 years when LTS versions change).
Non-LTS versions will be installed by the normal jobs as Composer installs the latest compatible version

See https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/.github/workflows/ci.yaml for an example setup (where I'm not skipping the definition of the env variable when `matrix.symfony` is `false` but I'm skipping the installation of Flex, which consumes it, producing an equivalent result)

PR #530, opened 2025-09-25
Remove incomplete service definitions

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 4.x
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

Inspired by https://github.com/symfony/monolog-bundle/pull/207#issuecomment-285097790.
These service definitions are unused and, to my understanding, were only kept for BC reasons. Maybe it's time to remove them?

### DIFF
--- a/config/monolog.php
+++ b/config/monolog.php
@@ -21,10 +21,8 @@ use Monolog\Formatter\LogstashFormatter;
 use Monolog\Formatter\NormalizerFormatter;
 use Monolog\Formatter\ScalarFormatter;
 use Monolog\Formatter\WildfireFormatter;
-use Monolog\Handler\FingersCrossed\ErrorLevelActivationStrategy;
 use Monolog\Logger;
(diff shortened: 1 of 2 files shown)


config/monolog.php
```diff
@@ -43,9 +41,6 @@
             ->args([abstract_arg('channel')])
             ->abstract()
 
-        ->set('monolog.activation_strategy.not_found', NotFoundActivationStrategy::class)->abstract()
-        ->set('monolog.handler.fingers_crossed.error_level_activation_strategy', ErrorLevelActivationStrategy::class)->abstract()
```
GromNaN, line 48: I believe that these definitions should first be marked as deprecated in 3.x

HypeMC: Yes, for some reason I thought they already were, see #531.

GromNaN:
Thank you @HypeMC.

PR #536, opened 2025-09-26
use PHPUnit 12.3 when possible

xabbuh:
| Q             | A
| ------------- | ---
| Branch?       | 4.x
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | 
| License       | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "symfony/polyfill-php84": "^1.30"
     },
     "require-dev": {
-        "phpunit/phpunit": "^11.5.41",
+        "phpunit/phpunit": "^11.5.41 || ^12.3",
         "symfony/console": "^7.3 || ^8.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -26,7 +26,7 @@
         "symfony/polyfill-php84": "^1.30"
     },
     "require-dev": {
-        "phpunit/phpunit": "^11.5.41",
+        "phpunit/phpunit": "^12.3",
```
derrabus: That won't work on PHP 8.2, will it.

xabbuh: hm, indeed

xabbuh: but we can use it on the PHP 8.3/8.4 jobs

GromNaN (approved):

GromNaN:
Good catch, thanks @xabbuh.

PR #541, opened 2025-10-02
Mark XML configuration tests as legacy

GromNaN:
| Q             | A
| ------------- | ---
| Branch?       | 3.x
| Bug fix?      | no (test only)
| New feature?  | no
| Deprecations? | no (deprecated by Symfony itself)
| Issues        | Related to https://github.com/symfony/symfony/pull/60568
| License       | MIT

Hide deprecation notices on Symfony 7.4 without asserting the deprecation message: 

```
  13x: Since symfony/dependency-injection 7.4: XML configuration format is deprecated, use YAML or PHP instead.
    1x in XmlMonologExtensionTest::testLoadWithSeveralHandlers from Symfony\Bundle\MonologBundle\Tests\DependencyInjection
    1x in XmlMonologExtensionTest::testLoadWithOverwriting from Symfony\Bundle\MonologBundle\Tests\DependencyInjection
    1x in XmlMonologExtensionTest::testLoadWithNewAtEnd from Symfony\Bundle\MonologBundle\Tests\DependencyInjection
    1x in XmlMonologExtensionTest::testLoadWithNewAndPriority from Symfony\Bundle\MonologBundle\Tests\DependencyInjection
    1x in XmlMonologExtensionTest::testHandlersWithChannels from Symfony\Bundle\MonologBundle\Tests\DependencyInjection
    ...

Remaining indirect deprecation notices (3)

  3x: Since symfony/dependency-injection 7.4: XML configuration format is deprecated, use YAML or PHP instead.
    1x in XmlMonologExtensionTest::testLoadWithOverwriting from Symfony\Bundle\MonologBundle\Tests\DependencyInjection
    1x in XmlMonologExtensionTest::testLoadWithNewAtEnd from Symfony\Bundle\MonologBundle\Tests\DependencyInjection
    1x in XmlMonologExtensionTest::testLoadWithNewAndPriority from Symfony\Bundle\MonologBundle\Tests\DependencyInjection
```

PR #543, opened 2025-10-03
Skip XML config tests with Symfony 8.0

GromNaN:
| Q             | A
| ------------- | ---
| Branch?       | 4.x
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | Related to https://github.com/symfony/symfony/pull/60568
| License       | MIT

Follow-up:
- https://github.com/symfony/monolog-bundle/pull/541

As soon as we keep support for Symfony 7.4, we must keep the tests on XML config format.

### DIFF
--- a/tests/DependencyInjection/XmlMonologExtensionTest.php
+++ b/tests/DependencyInjection/XmlMonologExtensionTest.php
@@ -11,7 +11,7 @@
 
 namespace Symfony\Bundle\MonologBundle\Tests\DependencyInjection;
 
-use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\Attributes\IgnoreDeprecations;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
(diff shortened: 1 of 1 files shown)


tests/DependencyInjection/FixtureMonologExtensionTestCase.php
```diff
@@ -191,9 +192,11 @@ public function testPsr3MessageProcessingDisabledOnNullHandler()
 
     public function testHandlersV2()
     {
-        $this->getContainer('handlers');
+        $container = $this->getContainer('handlers');
 
-        $this->expectNotToPerformAssertions();
```
GromNaN: I improved this test because I initially added the `expectUserDeprecationMessage` to `XmlMonologExtensionTest::loadFixture`. But Symfony 7.3 does not trigger this deprecation and `#[IgnoreDeprecation]` is better.
Keeping the test update as it still covers the configuration.

stof: shouldn't this test improvement be done in the 3.x branch already and merged up ?

GromNaN: Yes: https://github.com/symfony/monolog-bundle/pull/544

HypeMC (approved):

PR #550, opened 2025-10-23
Deprecate "mongo" handler type in favor of new "mongodb" syntax

jmikola:
| Q             | A
| ------------- | ---
| Branch?       | 3.x 
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | yes
| Issues        | [PHPORM-398](https://jira.mongodb.org/browse/PHPORM-398)
| License       | MIT

Define a new "mongodb" handler type. It accepts an "id" reference like the old "mongo" type; however, "uri" instead of a single "host" and "port". The "uri" option is more flexible.

Additionally, the "username" and "password" options have been renamed and are no longer used to modify the connection string directly ("mongo" never applied URL encoding). Instead, the options are set in the URI options array, which does not require encoding. The "mongodb" never requires a password, as a username alone is valid for some auth mechanisms.

Lastly, a "monolog-bundle" app name is specified when the bundle constructs a MongoDB\Client instance for both "mongo" and "mongodb" handler syntax.

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -90,6 +90,17 @@ use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
  *   - [level]: level name or int value, defaults to DEBUG
  *   - [bubble]: bool, defaults to true
  *
+ * - mongodb:
+ *    - mongodb:
+ *       - id: optional if uri is given
+ *       - uri: MongoDB connection string, optional if id is given
+ *       - [username]: Username for database authentication
+ *       - [password]: Password for database authentication
+ *       - [database]: Database to which logs are written (not used for auth), defaults to "monolog"
+ *       - [collection]: Collection to which logs are written, defaults to "logs"
+ *    - [level]: level name or int value, defaults to DEBUG
+ *    - [bubble]: bool, defaults to true
+ *
  * - elastic_search:
  *   - elasticsearch:
  *      - id: optional if host is given
@@ -648,6 +659,7 @@ class Configuration implements ConfigurationInterface
 
         $this->addGelfSection($handlerNode);
         $this->addMongoSection($handlerNode);
+        $this->addMongoDBSection($handlerNode);
         $this->addElasticsearchSection($handlerNode);
         $this->addRedisSection($handlerNode);
         $this->addPredisSection($handlerNode);
@@ -879,7 +891,7 @@ class Configuration implements ConfigurationInterface
                     ->ifTrue(function ($v) {
                         return !isset($v['id']) && !isset($v['host']);
                     })
-                    ->thenInvalid('What must be set is either the host or the id.')
+                    ->thenInvalid('The "mongo" handler configuration requires either a service "id" or a connection "host".')
                     ->end()
                     ->validate()
                     ->ifTrue(function ($v) {
@@ -891,7 +903,43 @@ class Configuration implements ConfigurationInterface
             ->end()
             ->validate()
                 ->ifTrue(function ($v) { return 'mongo' === $v['type'] && !isset($v['mongo']); })
-                ->thenInvalid('The mongo configuration has to be specified to use a MongoHandler')
+                ->thenInvalid('The "mongo" configuration has to be specified to use a "mongo" handler type.')
+            ->end()
+        ;
+    }
+
+    private function addMongoDBSection(ArrayNodeDefinition $handlerNode)
+    {
+        $handlerNode
+            ->children()
+                ->arrayNode('mongodb')
+                    ->canBeUnset()
+                    ->beforeNormalization()
+                        ->ifString()
+                        ->then(function ($v) { return ['id' => $v]; })
+                    ->end()
+                    ->children()
+                        ->scalarNode('id')
+                            ->info('ID of a MongoDB\Client service')
+                            ->example('doctrine_mongodb.odm.logs_connection')
+                        ->end()
+                        ->scalarNode('uri')->end()
+                        ->scalarNode('username')->end()
+                        ->scalarNode('password')->end()
+                        ->scalarNode('database')->defaultValue('monolog')->end()
+                        ->scalarNode('collection')->defaultValue('logs')->end()
+                    ->end()
+                    ->validate()
+                        ->ifTrue(function ($v) {
+                            return !isset($v['id']) && !isset($v['uri']);
+                        })
+                        ->thenInvalid('The "mongodb" handler configuration requires either a service "id" or a connection "uri".')
+                    ->end()
+                ->end()
+            ->end()
+            ->validate()
+                ->ifTrue(function ($v) { return 'mongodb' === $v['type'] && !isset($v['mongodb']); })
+                ->thenInvalid('The "mongodb" configuration has to be specified to use a "mongodb" handler type.')
             ->end()
         ;
     }
--- a/src/DependencyInjection/MonologExtension.php
+++ b/src/DependencyInjection/MonologExtension.php
@@ -265,6 +265,12 @@ class MonologExtension extends Extension
                 break;
 
             case 'mongo':
+                trigger_deprecation('symfony/monolog-bundle', '3.11', 'The "mongo" handler type is deprecated in MonologBundle since version 3.11.0, use the "mongodb" type instead.');
+
+                if (!class_exists('MongoDB\Client')) {
+                    throw new \RuntimeException('The "mongo" handler requires the mongodb/mongodb package to be installed.');
+                }
+
                 if (isset($handler['mongo']['id'])) {
                     $client = new Reference($handler['mongo']['id']);
                 } else {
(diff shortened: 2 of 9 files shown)


jmikola:
The [Fabbot failure](https://github.com/symfony/monolog-bundle/actions/runs/18736075269/job/53442874847?pr=550) seems unrelated:

> Error: Test case methods should not have a return type. Please apply the patch below.

None of the new test methods introduced in this PR specify a return type.

@GromNaN did this not come up in your previous PR (https://github.com/symfony/monolog-bundle/pull/485)?

src/DependencyInjection/Configuration.php
```diff
@@ -891,7 +903,40 @@ private function addMongoSection(ArrayNodeDefinition $handlerNode)
...
+                        ->scalarNode('username')->end()
+                        ->scalarNode('password')->end()
+                        ->scalarNode('database')->defaultValue('monolog')->end()
+                        ->scalarNode('collection')->defaultValue('logs')->end()
+                    ->end()
+                    ->validate()
+                        ->ifTrue(function ($v) {
+                            return !isset($v['id']) && !isset($v['uri']);
+                        })
+                        ->thenInvalid('What must be set is either the uri or the id.')
```
GromNaN:
```suggestion
                        ->thenInvalid('The "mongodb" handler configuration requires a service "id" or a connection "uri".')
```

src/DependencyInjection/MonologExtension.php
```diff
@@ -292,6 +295,43 @@ private function buildHandler(ContainerBuilder $container, $name, array $handler
                 ]);
                 break;
 
+            case 'mongodb':
```
GromNaN, line 300: DX improvement: throw an actionable exception when the mongodb lib is not installed.


```suggestion
            case 'mongodb':
                if (!class_exists('MongoDB\Client')) {
                    throw new \RuntimeException('The MongoDB Client library is not available. Try running "composer require mongodb/mongodb".');
                }
```

src/DependencyInjection/Configuration.php
```diff
@@ -891,7 +903,40 @@ private function addMongoSection(ArrayNodeDefinition $handlerNode)
             ->end()
             ->validate()
                 ->ifTrue(function ($v) { return 'mongo' === $v['type'] && !isset($v['mongo']); })
-                ->thenInvalid('The mongo configuration has to be specified to use a MongoHandler')
+                ->thenInvalid('The mongo configuration has to be specified to use a MongoDBHandler')
```
GromNaN:
```suggestion
                ->thenInvalid('The "mongo" configuration has to be specified to use a "mongo" handler type')
```

src/DependencyInjection/Configuration.php
```diff
@@ -891,7 +903,40 @@ private function addMongoSection(ArrayNodeDefinition $handlerNode)
...
+                        ->ifTrue(function ($v) {
+                            return !isset($v['id']) && !isset($v['uri']);
+                        })
+                        ->thenInvalid('What must be set is either the uri or the id.')
+                    ->end()
+                ->end()
+            ->end()
+            ->validate()
+                ->ifTrue(function ($v) { return 'mongodb' === $v['type'] && !isset($v['mongodb']); })
+                ->thenInvalid('The mongodb configuration has to be specified to use a MongoDBHandler')
```
GromNaN:
```suggestion
                ->thenInvalid('The "mongodb" configuration has to be specified to use a "mongodb" handler type.')
```

src/DependencyInjection/MonologExtension.php
```diff
@@ -292,6 +295,43 @@ private function buildHandler(ContainerBuilder $container, $name, array $handler
...
+                    $client,
+                    $handler['mongodb']['database'],
+                    $handler['mongodb']['collection'],
+                    $handler['level'],
+                    $handler['bubble'],
+                ]);
+
+                if (empty($handler['formatter'])) {
+                    $formatter = new Definition('Monolog\Formatter\MongoDBFormatter');
+                    $formatter->setPublic(false);
```
GromNaN: It's an inline service, it can't be public, you can remove this statement.

src/DependencyInjection/MonologExtension.php
```diff
@@ -292,6 +295,43 @@ private function buildHandler(ContainerBuilder $container, $name, array $handler
...
+                    if (isset($handler['mongodb']['password'])) {
+                        $uriOptions['password'] = $handler['mongodb']['password'];
+                    }
+
+                    $client = new Definition('MongoDB\Client', [
+                        $handler['mongodb']['uri'],
+                        $uriOptions,
+                    ]);
+
+                    $client->setPublic(false);
```
GromNaN: It's an inline service, it can't be public, you can remove this statement.

GromNaN:
> The [Fabbot failure](https://github.com/symfony/monolog-bundle/actions/runs/18736075269/job/53442874847?pr=550) seems unrelated:
> 
> > Error: Test case methods should not have a return type. Please apply the patch below.
> 
> None of the new test methods introduced in this PR specify a return type.
> 
> @GromNaN did this not come up in your previous PR (#485)?

We added the rule after. We should probably remove the return type in this branch. It's annoying to ignore fabbot failures.

Edit: you can rebase.

src/DependencyInjection/MonologExtension.php
jmikola: I saw that Definition has `private bool $public = false`. I assume most of these `setPublic(false)` calls throughout MonologExtension are just obsolete.

src/DependencyInjection/Configuration.php
jmikola: I wasn't a fan of the previous error messages but kept the format for consistency. Happy to make these changes, though. Will make a corresponding change to the "mongo" configuration as well.

src/DependencyInjection/Configuration.php
```diff
@@ -891,7 +903,40 @@ private function addMongoSection(ArrayNodeDefinition $handlerNode)
...
+        $handlerNode
+            ->children()
+                ->arrayNode('mongodb')
+                    ->canBeUnset()
+                    ->beforeNormalization()
+                        ->ifString()
+                        ->then(function ($v) { return ['id' => $v]; })
+                    ->end()
+                    ->children()
+                        ->scalarNode('id')->end()
```
GromNaN: We can document the configuration:
```suggestion
                        ->scalarNode('id')
                            ->info('ID of a MongoDB\Client service')
                            ->example('doctrine_mongodb.odm.logs_connection')
                        ->end()
```

src/DependencyInjection/MonologExtension.php
```diff
@@ -292,6 +297,44 @@ private function buildHandler(ContainerBuilder $container, $name, array $handler
...
+                        $uriOptions['username'] = $handler['mongodb']['username'];
+                    }
+
+                    if (isset($handler['mongodb']['password'])) {
+                        $uriOptions['password'] = $handler['mongodb']['password'];
+                    }
+
+                    $client = new Definition('MongoDB\Client', [
+                        $handler['mongodb']['uri'],
+                        $uriOptions,
```
GromNaN, line 320: Should we have `writeConcern: 0 or 1` by default for logs?
You can accept driver options.

GromNaN (approved):
A few ideas if you want to refine it, but otherwise it looks good to me!

GromNaN:
For the error in the CI, we can add `extension: mongodb` to [setup-php](https://github.com/symfony/monolog-bundle/blob/d3cf193260a8bb0884e6cff521e386ff8927e9bb/.github/workflows/ci.yaml#L39).

src/DependencyInjection/Configuration.php
jmikola: Ah, I love that we can add inline documentation like this. That will be super useful for [PHPLIB-1741](https://jira.mongodb.org/browse/PHPLIB-1741).

src/DependencyInjection/MonologExtension.php
jmikola: That can be specified in the `uri` option. I only wanted to break out `username` and `password` since (a) they might come from separate env vars and (b) they were supported in the original `mongo` syntax and I wanted to avoid dealing with URI encoding.

We can revisit when adding more options across the board.

> You can accept driver options

Were you actually referring to `$driverOptions` (i.e. third ctor param) or just URI options?

tests/DependencyInjection/MonologExtensionTest.php
```diff
@@ -932,6 +933,164 @@ public function testElasticsearchAndElasticaHandlers()
         $this->assertSame(['hosts' => ['es:9200'], 'transport' => 'Http'], $elasticaClient->getArgument(0));
     }
 
+    /** @group legacy */
+    public function testMongo()
+    {
+        if (!class_exists('MongoDB\Client') {
```
GromNaN:
```suggestion
        if (!class_exists('MongoDB\Client')) {
```

GromNaN (approved):

GromNaN:
Thank you @jmikola.

PR #553, opened 2025-10-27
Remove support for the `DebugHandler`

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 3.x
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

The `DebugHandler` was [removed way back in Symfony 4.0](https://github.com/symfony/symfony/pull/20416). Since the bundle only supports >=6.4, this is no longer needed.

This was an oversight in #526.

PR #554, opened 2025-10-27
Use `::class` constant

HypeMC:
| Q             | A
| ------------- | ---
| Branch?       | 4.x
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

Currently, we have a mix of strings and constants. I suggest using constants everywhere. This actually helped me spot #553.

I targeted 4.x because there would be a lot of conflicts if done on 3.x.

Also, I intentionally didn't import non-Symfony classes, as I think this is more readable. That's just my opinion though, I'm fine with imports as well.

### DIFF
--- a/tests/DependencyInjection/MonologExtensionTest.php
+++ b/tests/DependencyInjection/MonologExtensionTest.php
@@ -46,7 +46,7 @@ class MonologExtensionTest extends DependencyInjectionTestCase
         $this->assertDICDefinitionMethodCallAt(1, $logger, 'pushHandler', [new Reference('monolog.handler.main')]);
 
         $handler = $container->getDefinition('monolog.handler.main');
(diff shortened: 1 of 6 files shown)


tests/DependencyInjection/MonologExtensionTest.php
```diff
@@ -444,9 +444,6 @@ public static function provideLoglevelParameterConfig(): array
 
     public function testProcessorAutoConfiguration()
     {
-        if (!interface_exists('Monolog\ResettableInterface')) {
-            $this->markTestSkipped('The ResettableInterface is not available.');
-        }
```
HypeMC, line 452: Another thing I spotted while doing this.

GromNaN (approved):

GromNaN:
Thank you @HypeMC.

PR #555, opened 2025-10-27
Remove "mongo" handler type

GromNaN:
| Q             | A
| ------------- | ---
| Branch?       | 4.x
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | After https://github.com/symfony/monolog-bundle/pull/550
| License       | MIT

The `mongo` handler type is replaced by `mongodb`.

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -79,18 +79,6 @@ use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
  *   - [filename_format]: string, defaults to '{filename}-{date}'
  *   - [date_format]: string, defaults to 'Y-m-d'
  *
- * - mongo:
- *   - mongo:
- *      - id: optional if host is given
- *      - host: database host name, optional if id is given
- *      - [port]: defaults to 27017
- *      - [user]: database user name
- *      - pass: mandatory only if user is present
- *      - [database]: defaults to monolog
- *      - [collection]: defaults to logs
- *   - [level]: level name or int value, defaults to DEBUG
- *   - [bubble]: bool, defaults to true
- *
  * - mongodb:
  *    - mongodb:
  *       - id: optional if uri is given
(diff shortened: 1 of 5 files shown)


HypeMC (approved):
Fabbot failure should be fixed.

src/DependencyInjection/Configuration.php
```diff
@@ -584,7 +584,6 @@ public function getConfigTreeBuilder(): TreeBuilder
             ->end();
 
         $this->addGelfSection($handlerNode);
-        $this->addMongoSection($handlerNode);
```
HypeMC, line 587: The [comments](https://github.com/symfony/monolog-bundle/blob/27d779a384e4da1e4132e7a427414bbc1b0a8217/src/DependencyInjection/Configuration.php#L82) on the top should be removed as well.

HypeMC: Also from the XSD.

GromNaN: Thank you for catching what I had missed. It 's fixed.

HypeMC (approved):

src/DependencyInjection/Configuration.php
HypeMC: Just noticed, the [comments](https://github.com/symfony/monolog-bundle/blob/27d779a384e4da1e4132e7a427414bbc1b0a8217/src/DependencyInjection/Configuration.php#L82) part is still there.

src/DependencyInjection/MonologExtension.php
```diff
@@ -302,10 +269,6 @@ private function buildHandler(ContainerBuilder $container, string $name, array $
                 }
                 break;
 
-            case 'elasticsearch':
```
GromNaN, line 305: [`elasticsearch` case reappeared during the upmerge](https://github.com/symfony/monolog-bundle/commit/c2c37a99a3197c890a853981574e55355ac30bf3#diff-40ecc5cc3a85f809629efc60aa68e5e437027439d3778fde3874ba2d04f80a7cR303-R305).

PR #557, opened 2025-11-04
Added the handler_class option to customize a handler while keeping i…

charliefancelli:
…ts original type.

| Q             | A
| ------------- | ---
| Branch?       | 3.x 
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #... 
| License       | MIT

This pull request adds support for specifying a custom handler class for Monolog handlers via configuration. This allows users to define which PHP class should be used for a handler, increasing flexibility and extensibility in logging configuration. The main changes include schema and config updates, as well as logic to validate and use the custom handler class if provided.

**Configuration and Schema Updates:**
* Added a new `handler_class` attribute to the Monolog XML schema (`monolog-1.0.xsd`) to allow specifying a custom handler class in XML configuration.
* Added a new `handler_class` option to the handler configuration tree in `Configuration.php`, defaulting to `null`.

**Handler Instantiation Logic:**
* Updated the handler building logic in `MonologExtension.php` to use the specified `handler_class` if provided, including a runtime check that the class exists, otherwise falling back to the default handler class for the given type.

**Sample**
```
monolog:
    handlers:
        main:
            type: stream
            handler_class: App\Monolog\CustomStreamHandler
            path: '%kernel.logs_dir%/%kernel.environment%.log'
```

### DIFF
--- a/config/schema/monolog-1.0.xsd
+++ b/config/schema/monolog-1.0.xsd
@@ -95,6 +95,7 @@
         <xsd:attribute name="slack-team" type="xsd:string" />
         <xsd:attribute name="region" type="xsd:string" />
         <xsd:attribute name="enabled" type="xsd:boolean" />
+        <xsd:attribute name="handler_class" type="xsd:string" />
     </xsd:complexType>
 
     <xsd:simpleType name="level">
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -661,6 +661,7 @@ class Configuration implements ConfigurationInterface
                 ->end()
                 ->scalarNode('formatter')->end()
                 ->booleanNode('nested')->defaultFalse()->end()
+                ->scalarNode('handler_class')->defaultNull()->end()
             ->end();
 
         $this->addGelfSection($handlerNode);
--- a/src/DependencyInjection/MonologExtension.php
+++ b/src/DependencyInjection/MonologExtension.php
@@ -169,7 +169,7 @@ class MonologExtension extends Extension
         if (null !== $handler['handler_class']) {
             $handlerClass = $handler['handler_class'];
             if (!class_exists($handlerClass)) {
-                throw new \RuntimeException(sprintf('The handler class "%s" does not exist.', $handlerClass));
+                throw new \RuntimeException(\sprintf('The handler class "%s" does not exist.', $handlerClass));
             }
         } else {
             $handlerClass = $this->getHandlerClassByType($handler['type']);
--- a/src/DependencyInjection/MonologExtension.php
+++ b/src/DependencyInjection/MonologExtension.php
@@ -166,7 +166,15 @@ class MonologExtension extends Extension
             return $handlerId;
         }
 
-        $handlerClass = $this->getHandlerClassByType($handler['type']);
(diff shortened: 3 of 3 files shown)


GromNaN:
You can already specify a `type: service, id: <service_id>`. I think that's the way.

charliefancelli:
Defining a service requires redefining all the handler options at the service level, whereas this approach allows you to keep the original handler options and handle your own logic directly within the handler.
This way, you can switch from a “native” handler to a “custom” one simply by adding the option.

stof:
This also means that we cannot ever change our handling of known handler types in the future because any change in the structure of the definition becomes a potential BC breaks for custom classes. That's a maintenance nightmare.

Supporting custom handlers only through custom service definitions was an explicit choice to make it possible to keep the bundle evolving without having to release every bugfix as a major version due to the BC impact.

So a big -1 vote from me on this PR.

charliefancelli:
I understand your concern, and indeed this feature requires good knowledge of the codebase and proper maintenance awareness.
When you choose to use it, you implicitly know that updating the bundle will require some attention.
In my view, the bundle should allow developers to use this feature knowingly, but not take responsibility for supporting it over time.

Moreover, I don’t really see how this differs from using a service decoration, which is a common pattern in Symfony.
Once you decorate a service, you already accept that any future update may introduce potential issues with the decorator.

GromNaN:
We removed the `.class` parameters in Symfony and 3rd party bundles because that was adding maintenance complexity and unexpected breaking changes. A service definition is a whole: class + arguments + method calls... As the developer that change the class of a service, you understand what that implies, but for other developers that will have to update the dependencies of the project, that will end in a nightmare.

What do you want to customize in `CustomStreamHandler`? Maybe that's as feature that would benefit to Monolog or this bundle.

charliefancelli:
@GromNaN I’ve been wondering for years why Monolog doesn’t natively support log compression during rotation.
As far as I remember, this topic was brought up a long time ago on the Monolog side, but it was rejected (I never really understood why, though).

Since we have a bundle that gives us an additional layer on top, I was able to handle this directly there while keeping the entire Monolog configuration intact.
Sure, creating a service would be possible, but when you have 30 different loggers with different setups, you don’t want to redefine 30 services just to manage that. I simply wanted to hook into the rotation function to handle compression.

This way, we can switch from “normal” mode to “compressed” mode just by modifying the handler.

That said, I completely understand the maintainability concerns

stof:
> Moreover, I don’t really see how this differs from using a service decoration, which is a common pattern in Symfony.
> Once you decorate a service, you already accept that any future update may introduce potential issues with the decorator.

That's not true. When using service decoration, your own service definition is responsible for the full configuration of the instance of your class, while the core service definition is responsible for the configuration of the instance of the core class. That's makes it a lot less fragile regarding BC.

GromNaN:
> I’ve been wondering for years why Monolog doesn’t natively support log compression during rotation. As far as I remember, this topic was brought up a long time ago on the Monolog side, but it was rejected (I never really understood why, though).

Indeed, that was requested in https://github.com/Seldaek/monolog/issues/1202 and rejected.

Enabling custom handler types and configuration would be a solution: https://github.com/symfony/monolog-bundle/issues/558

PR #559, opened 2025-11-17
Fix `rollbar` handler to use `RollbarLogger` with Monolog 2+

websirnik:
| Q             | A
| ------------- | ---
| Branch?       | 3.x 
| Bug fix?      | yes
| New feature?  | o
| Deprecations? | no
| License       | MIT

Improves the compatibility with the latest Rollbar library. Rollbar has renamed their classes a while ago.

### DIFF
--- a/src/DependencyInjection/MonologExtension.php
+++ b/src/DependencyInjection/MonologExtension.php
@@ -930,10 +930,10 @@ class MonologExtension extends Extension
                 } else {
                     $config = $handler['config'] ?: [];
                     $config['access_token'] = $handler['token'];
-                    $rollbar = new Definition('RollbarNotifier', [
+                    $rollbar = new Definition('Rollbar\RollbarLogger', [
                         $config,
                     ]);
-                    $rollbarId = 'monolog.rollbar.notifier.'.sha1(json_encode($config));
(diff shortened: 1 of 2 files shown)


src/DependencyInjection/MonologExtension.php
```diff
@@ -930,10 +930,10 @@ private function buildHandler(ContainerBuilder $container, $name, array $handler
                 } else {
                     $config = $handler['config'] ?: [];
                     $config['access_token'] = $handler['token'];
-                    $rollbar = new Definition('RollbarNotifier', [
+                    $rollbar = new Definition('Rollbar\RollbarLogger', [
```
GromNaN: The signature of the `RollbarHandler ` was modified in Monolog 2.0.0 by https://github.com/Seldaek/monolog/pull/1042. This class must be set conditionally depending on the Monolog version.

In MonologBundle v4, since Monolog 3 is required, we can make this change without condition.

GromNaN (changes requested):

GromNaN (approved):

GromNaN:
Thanks for fixing this bug @websirnik.

PR #560, opened 2025-11-26
Make each handler type have its own extension class

### LINKED ISSUE #558: Make each handler type have its own configuration and enable custom handler types
For the moment, the handler configuration options are all mixed together. This means that it is necessary to validate the options in the extension, even though this is the role of the config tree.

With the introduction of static configuration validation by array shape (and JSON schema?), it would be beneficial for each handler type to have its own configuration. This is already the case for `mailer`, `redis`, `mongodb`, and `elasticsearch`. We could generalize this with one class per handler, having two methods: one that create the nested config tree and the other that create the handler service definition.

With this approach, we can enable bundles/application to register additional handler types with their own config logic.

This requires a massive deprecation of the existing configuration, we might have some permanent mapping logic to pre-process the configuration.

Inspiration: https://github.com/thephpleague/flysystem-bundle/pull/186

GromNaN:
| Q             | A
| ------------- | ---
| Branch?       | 3.x
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no (not yet)
| Issues        | Fix #558
| License       | MIT

This is a partial PR to gather feedbacks.
- Each Monolog handler has its own configuration and extension class
- Some handlers can share the same with different properties (name, handler class)
- They are all registered in the MonologExtension
- 3rd party bundles can add their own handler extension classes


src/DependencyInjection/Handler/HandlerExtensionInterface.php
```diff
@@ -0,0 +1,38 @@
...
+     */
+    public function buildArrayNode(ArrayNodeDefinition $handlerNode): void;
+
+    /**
+     * @param array{
+     *     level: string|int,
+     *     bubble: bool,
+     *     formatter: string|null,
+     * } $config Generic options
+     * @param array{} $handler Specific handler options
```
stof, line 35: this is wrong. `array{}` is the shape representing an empty array.

stof (commented):
I don't think we want external bundles to extend the MonologBundle configuration class. This creates extra complexity:

- depending on where the Configuration class is accessed, the extensions might be registered or no (we have that issue in SecurityBundle, which is the only bundle I'm aware of doing such thing
- it makes it harder to ensure BC if we don't control the config tree
- most of our handlers supported in core don't group their config settings in a dedicated sub-node in the config tree, which means they cannot be migrated to this new system (at least not without a complex BC layer to deprecate the existing nodes, which will be very hard to implement properly given that some of our config settings are used by _multiple_ handlers).

We already have the `service` type to allow configuring handlers that are not supported in core.

src/DependencyInjection/Handler/HandlerExtensionInterface.php
GromNaN: This is on purpose. By default this is an empty array. You have to specify another array shape in subclasses.

stof: That's not the proper way to describe such API. We would need to use a generic type with `THandlerConfig of array<string, mixed>`, with child classes describing their array shape as the template type.

mbabker:
> * depending on where the Configuration class is accessed, the extensions might be registered or no (we have that issue in SecurityBundle, which is the only bundle I'm aware of doing such thing

`knplabs/knp-gaufrette-bundle`, `liip/imagine-bundle`, `payum/payum-bundle`, and `sylius/theme-bundle` all do something similar

ThomasLandauer:
Could this also be used to pass the handler's FQCN as `type`, rather than some string?:
```php
'handlers' => [
    'foobar' => [
        'type' => StreamHandler::class
```

PR #566, opened 2025-12-08
Add info to the `monolog.channels` config

### LINKED ISSUE #565: Suggestion: Config: Rename `channels` to `custom_channels`
What I mean:
```php
return App::config([
    'monolog' => [
        'custom_channels' => [ // <= new name
            // ...
```
Why?
It took me forever to finally understand that whatever you pass there, will never be a *full* list of all channels (because of Symfony's built-in channels like `router` or `security`).
Renaming this to *custom*_channels would nicely convey the message that more channels exist.

But it would be a BC break.

GromNaN:
| Q             | A
| ------------- | ---
| Branch?       | 3.x
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #565
| License       | MIT

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -416,6 +416,7 @@ class Configuration implements ConfigurationInterface
             ->children()
                 ->scalarNode('use_microseconds')->defaultTrue()->end()
                 ->arrayNode('channels')
+                    ->info('List of additional channels to create. The "app" channel is already created by default. Channels are also automatically created for services using the "monolog.logger" DI tag with a custom channel attribute.')
                     ->canBeUnset()
                     ->prototype('scalar')->end()
                 ->end()


src/DependencyInjection/Configuration.php
```diff
@@ -416,6 +416,7 @@ public function getConfigTreeBuilder(): TreeBuilder
             ->children()
                 ->scalarNode('use_microseconds')->defaultTrue()->end()
                 ->arrayNode('channels')
+                    ->info('List of additional channels to create. The "app" channel is already created by default.')
```
derrabus: Should we also mention the relation to the DI tag here?

stof: We should indeed probably mention that additional channels are also created based on their usages in services.

derrabus: Yes, that's what I meant.

GromNaN: Done, I've updated the message to mention the `monolog.logger` DI tag.

HypeMC (approved):

PR #569, opened 2026-01-20
Import Extension from the DependencyInjection component rather than HttpKernel

mickverm:
| Q             | A
| ------------- | ---
| Branch?       | 4.x
| Bug fix?      | no
| New feature?  | no
| Deprecations? | no
| Issues        | -
| License       | MIT

The Extension class from the HttpKernel component is marked as internal since Symfony 7.1 and is going to be deprecated in 8.1 (see https://github.com/symfony/symfony/pull/62434). 

I can target 3.x if preferred.


GromNaN (approved):

GromNaN:
> I can target 3.x if preferred.

The 3.x branch will not support Symfony 8.0, so there is no need to backport this change.

mickverm:
> The 3.x branch will not support Symfony 8.0, so there is no need to backport this change.

That's what I thought, the only reason to backport would be to circumvent the `@internal` on the HttpKernel Extension.

xabbuh (approved):

GromNaN:
Thank you @mickverm.

PR #577, opened 2026-04-07
Add `base_path` option for `LineFormatter`

### LINKED ISSUE #478: Add support for setting the basePath on line formatter
See https://github.com/Seldaek/monolog/releases/tag/3.6.0 / https://github.com/Seldaek/monolog/pull/1873

It would be neat I think if we could do `base_path: %kernel.project_dir%` on a handler and it would configure the line formatter for it with base path set.

It could work like the `include_stacktraces` functionality I'd say.

Amoifr:
| Q             | A
| ------------- | ---
| Branch?       | 4.x
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #478
| License       | MIT

This adds the `base_path` configuration option for handlers, allowing to strip a base path from stack traces. Uses Monolog 3.6.0's `LineFormatter::setBasePath()`.

The existing `MonologBundle::includeStacktraces()` method is replaced by a `FormatterConfigurator` service that handles both options.

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -36,6 +36,7 @@ use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
  *   - [bubble]: bool, defaults to true
  *   - [file_permission]: int|null, defaults to null (0644)
  *   - [use_locking]: bool, defaults to false
+ *   - [base_path]: string|null, base path to strip from file paths in stack traces (e.g. %kernel.project_dir%)
  *
  * - console:
  *   - [verbosity_levels]: level => verbosity configuration
@@ -416,6 +417,7 @@ final class Configuration implements ConfigurationInterface
                 ->booleanNode('interactive_only')->defaultFalse()->end()
                 ->scalarNode('app_name')->defaultNull()->end()
                 ->booleanNode('include_stacktraces')->defaultFalse()->end()
+                ->scalarNode('base_path')->defaultNull()->end()
                 ->arrayNode('process_psr_3_messages')
                     ->addDefaultsIfNotSet()
                     ->beforeNormalization()
--- a/src/DependencyInjection/FormatterConfigurator.php
+++ b/src/DependencyInjection/FormatterConfigurator.php
@@ -1,0 +1,45 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MonologBundle\DependencyInjection;
+
+use Monolog\Formatter\JsonFormatter;
+use Monolog\Formatter\LineFormatter;
+use Monolog\Handler\HandlerInterface;
+
+/**
+ * Configures formatter options on handlers.
+ *
+ * @author Pascal Cescon <pascal.cescon@gmail.com>
+ *
+ * @internal
+ */
+final class FormatterConfigurator
+{
+    public function __construct(
+        private bool $includeStacktraces = false,
+        private ?string $basePath = null,
+    ) {
+    }
+
+    public function __invoke(HandlerInterface $handler): void
+    {
+        $formatter = $handler->getFormatter();
+
+        if ($this->includeStacktraces && ($formatter instanceof LineFormatter || $formatter instanceof JsonFormatter)) {
+            $formatter->includeStacktraces();
+        }
+
+        if (null !== $this->basePath && $formatter instanceof LineFormatter) {
+            $formatter->setBasePath($this->basePath);
+        }
+    }
+}
--- a/src/DependencyInjection/MonologExtension.php
+++ b/src/DependencyInjection/MonologExtension.php
@@ -20,7 +20,6 @@ use Monolog\Processor\PsrLogMessageProcessor;
 use Monolog\ResettableInterface;
 use Symfony\Bridge\Monolog\Handler\FingersCrossed\HttpCodeActivationStrategy;
 use Symfony\Bridge\Monolog\Processor\TokenProcessor;
-use Symfony\Bundle\MonologBundle\MonologBundle;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\Argument\BoundArgument;
 use Symfony\Component\DependencyInjection\ChildDefinition;
@@ -147,8 +146,12 @@ final class MonologExtension extends Extension
         $handlerClass = $this->getHandlerClassByType($handler['type']);
         $definition = new Definition($handlerClass);
 
-        if ($handler['include_stacktraces']) {
-            $definition->setConfigurator([MonologBundle::class, 'includeStacktraces']);
+        if ($handler['include_stacktraces'] || null !== $handler['base_path']) {
+            $configurator = new Definition(FormatterConfigurator::class, [
+                $handler['include_stacktraces'],
+                $handler['base_path'],
+            ]);
+            $definition->setConfigurator([$configurator, '__invoke']);
         }
 
         if (null === $handler['process_psr_3_messages']['enabled']) {
--- a/tests/DependencyInjection/FormatterConfiguratorTest.php
+++ b/tests/DependencyInjection/FormatterConfiguratorTest.php
@@ -1,0 +1,127 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Bundle\MonologBundle\Tests\DependencyInjection;
+
+use Monolog\Formatter\JsonFormatter;
+use Monolog\Formatter\LineFormatter;
+use Monolog\Handler\StreamHandler;
+use Monolog\LogRecord;
+use PHPUnit\Framework\TestCase;
+use Symfony\Bundle\MonologBundle\DependencyInjection\FormatterConfigurator;
+
+class FormatterConfiguratorTest extends TestCase
+{
+    public function testIncludeStacktracesOnLineFormatter()
+    {
+        $handler = new StreamHandler('php://memory');
+        $formatter = new LineFormatter();
+        $handler->setFormatter($formatter);
+
+        $configurator = new FormatterConfigurator(includeStacktraces: true);
+        $configurator($handler);
+
+        // Verify includeStacktraces was called by formatting a record with an exception
+        $exception = new \Exception('Test exception');
+        $record = new LogRecord(
+            datetime: new \DateTimeImmutable(),
+            channel: 'test',
+            level: \Monolog\Level::Error,
+            message: 'Test',
+            context: ['exception' => $exception],
+        );
+        $output = $handler->getFormatter()->format($record);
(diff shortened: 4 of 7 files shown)


src/DependencyInjection/MonologExtension.php
```diff
@@ -147,8 +148,15 @@ private function buildHandler(ContainerBuilder $container, string $name, array $
         $handlerClass = $this->getHandlerClassByType($handler['type']);
         $definition = new Definition($handlerClass);
 
-        if ($handler['include_stacktraces']) {
-            $definition->setConfigurator([MonologBundle::class, 'includeStacktraces']);
```
stof, line 151: is this `MonologBundle::includeStacktraces` method still used ? If no, it should be removed

src/DependencyInjection/MonologExtension.php
```diff
@@ -147,8 +148,15 @@ private function buildHandler(ContainerBuilder $container, string $name, array $
...
+        if ($handler['include_stacktraces'] || null !== $handler['base_path']) {
+            $configurator = new Definition(FormatterConfigurator::class, [
+                $handler['include_stacktraces'],
+                $handler['base_path'],
+            ]);
+            $configurator->setPublic(false);
+            $configuratorId = \sprintf('monolog.handler.%s.formatter_configurator', $name);
```
stof: this naming pattern could cause issues with handlers than have a `.` in their name, as it might reuse the same id for 2 different things.
I suggest name it `monolog.formatter_configurator.%s` instead.

tests/DependencyInjection/FormatterConfiguratorTest.php
```diff
@@ -0,0 +1,95 @@
...
+    {
+        $handler = new StreamHandler('php://memory');
+        $formatter = new LineFormatter();
+        $handler->setFormatter($formatter);
+
+        $configurator = new FormatterConfigurator(includeStacktraces: true);
+        $configurator($handler);
+
+        // LineFormatter::includeStacktraces() was called - we verify by checking the formatter still works
+        $this->assertInstanceOf(LineFormatter::class, $handler->getFormatter());
```
stof: this does not actually verify that `LineFormatter::includeStacktraces()` was called.

Making the FormatterConfigurator a no-op would still make this test pass.

src/DependencyInjection/FormatterConfigurator.php
```diff
@@ -0,0 +1,45 @@
...
+
+        if ($this->includeStacktraces && ($formatter instanceof LineFormatter || $formatter instanceof JsonFormatter)) {
+            $formatter->includeStacktraces();
+        }
+
+        if (null !== $this->basePath && $formatter instanceof LineFormatter) {
+            $formatter->setBasePath($this->basePath);
+        }
+    }
+}
```
stof: please ensure that the file ends with a EOL

stof (changes requested):

src/DependencyInjection/MonologExtension.php
Amoifr: Done, removed the method and unused imports.

src/DependencyInjection/MonologExtension.php
Amoifr: Done, renamed to `monolog.formatter_configurator.%s`.

tests/DependencyInjection/FormatterConfiguratorTest.php
Amoifr: Done, tests now format a record with an exception and verify the output contains `[stacktrace]` (LineFormatter) or `"trace":` (JsonFormatter).

src/DependencyInjection/FormatterConfigurator.php
Amoifr: Done!

src/DependencyInjection/MonologExtension.php
```diff
@@ -147,8 +148,15 @@ private function buildHandler(ContainerBuilder $container, string $name, array $
...
         $definition = new Definition($handlerClass);
 
-        if ($handler['include_stacktraces']) {
-            $definition->setConfigurator([MonologBundle::class, 'includeStacktraces']);
+        if ($handler['include_stacktraces'] || null !== $handler['base_path']) {
+            $configurator = new Definition(FormatterConfigurator::class, [
+                $handler['include_stacktraces'],
+                $handler['base_path'],
+            ]);
+            $configurator->setPublic(false);
```
GromNaN: If possible, it would be better if the configurator remained private (and inlined).

GromNaN (commented):
Thank you for your help.

src/DependencyInjection/MonologExtension.php
```diff
@@ -20,6 +20,7 @@
 use Monolog\ResettableInterface;
 use Symfony\Bridge\Monolog\Handler\FingersCrossed\HttpCodeActivationStrategy;
 use Symfony\Bridge\Monolog\Processor\TokenProcessor;
+use Symfony\Bundle\MonologBundle\DependencyInjection\FormatterConfigurator;
 use Symfony\Bundle\MonologBundle\MonologBundle;
```
HypeMC: Fabbot is right about these two imports, they're not needed.

src/DependencyInjection/Configuration.php
```diff
@@ -416,6 +417,7 @@ public function getConfigTreeBuilder(): TreeBuilder
                 ->booleanNode('interactive_only')->defaultFalse()->end()
                 ->scalarNode('app_name')->defaultNull()->end()
                 ->booleanNode('include_stacktraces')->defaultFalse()->end()
+                ->scalarNode('base_path')->defaultNull()->end()
```
HypeMC, line 420: Lowest deps are currently failing. Maybe add some validation for Monolog <3.6, eg:
https://github.com/symfony/monolog-bundle/blob/c012c6aba13129eb02aa7dd61e66e720911d8598/src/DependencyInjection/Configuration.php#L698-L701

and update the tests, or bump the min version to 3.6?

src/DependencyInjection/MonologExtension.php
Amoifr: Done! Removed the explicit `setPublic(false)` (private by default) and switched to `[Reference, '__invoke']` syntax. I couldn't fully inline the Definition into `setConfigurator()` as it only accepts `Reference|array|string|null`, but the service stays private and scoped per handler.

src/DependencyInjection/MonologExtension.php
Amoifr: Done, good catch! Removed both unused imports.

src/DependencyInjection/Configuration.php
Amoifr: Done! Added a version validation for `base_path` requiring monolog/monolog >=3.6, following the same pattern as the `interactive_only` validation.

Amoifr:
Thanks @GromNaN and @HypeMC for the thorough review! All feedback has been addressed in the latest push. 🙏

src/DependencyInjection/Configuration.php
```diff
@@ -695,6 +697,10 @@ public function getConfigTreeBuilder(): TreeBuilder
                 ->ifTrue(function ($v) { return 'server_log' === $v['type'] && empty($v['host']); })
                 ->thenInvalid('The host has to be specified to use a ServerLogHandler')
             ->end()
+            ->validate()
+                ->ifTrue(function ($v) { return null !== $v['base_path'] && version_compare(InstalledVersions::getVersion('monolog/monolog'), '3.6.0', '<'); })
```
GromNaN: I think we prefer to use feature detection rather than version checking with Composer. Also adding the `static function` as CS fix.
```suggestion
                ->ifTrue(static function ($v) { return null !== $v['base_path'] && method_exists(\Monolog\Formatter\LineFormatter::class, 'setBasePath') })
```
/cc @HypeMC

src/DependencyInjection/MonologExtension.php
```diff
@@ -147,8 +146,14 @@ private function buildHandler(ContainerBuilder $container, string $name, array $
...
-        if ($handler['include_stacktraces']) {
-            $definition->setConfigurator([MonologBundle::class, 'includeStacktraces']);
+        if ($handler['include_stacktraces'] || null !== $handler['base_path']) {
+            $configurator = new Definition(FormatterConfigurator::class, [
+                $handler['include_stacktraces'],
+                $handler['base_path'],
+            ]);
+            $configuratorId = \sprintf('monolog.formatter_configurator.%s', $name);
+            $container->setDefinition($configuratorId, $configurator);
+            $definition->setConfigurator([new Reference($configuratorId), '__invoke']);
```
GromNaN: It seems to me that inlining works like this:
```suggestion
            $definition->setConfigurator([$configurator, '__invoke']);
```

GromNaN (commented):
Could you rebase on 4.x to fix CS issues?

src/DependencyInjection/MonologExtension.php
Amoifr: Done! Inlined the Definition directly as `[$configurator, '__invoke']` — no more service registration. Had to update the tests accordingly since the configurator is now retrieved from the handler definition via `getConfigurator()` instead of from the container.

Also rebased on 4.x as requested.

src/DependencyInjection/Configuration.php
Amoifr: Good call! Switched to `method_exists(\Monolog\Formatter\LineFormatter::class, 'setBasePath')` and added the `static function` CS fix.

src/DependencyInjection/Configuration.php
```diff
@@ -695,6 +697,10 @@ public function getConfigTreeBuilder(): TreeBuilder
                 ->ifTrue(static function ($v) { return 'server_log' === $v['type'] && empty($v['host']); })
                 ->thenInvalid('The host has to be specified to use a ServerLogHandler')
             ->end()
+            ->validate()
+                ->ifTrue(static function ($v) { return null !== $v['base_path'] && !method_exists(\Monolog\Formatter\LineFormatter::class, 'setBasePath'); })
```
stof: A simpler implementation would be to bump the composer dependency from `^3.5` to `^3.6` so that the feature is always available. The latest monolog release is 3.10 anyway. There is no reason for a project to try to upgrade monolog-bundle while sticking to monolog 3.5.0.

GromNaN: Definitely the best solution. Thanks for thinking of that, @stof

src/DependencyInjection/MonologExtension.php
GromNaN: They still need to be removed.

src/DependencyInjection/Configuration.php
Amoifr: Makes total sense! Bumped the composer dependency to `^3.6` and removed the validation entirely. Much cleaner, thanks @stof!

src/DependencyInjection/MonologExtension.php
Amoifr: Oops, they came back with the rebase! Removed again.

Amoifr:
All feedback addressed in the latest push:
- Bumped monolog dependency to `^3.6` instead of runtime validation (thanks @stof!)
- Configurator fully inlined as `[$definition, '__invoke']` (thanks @GromNaN!)
- Removed the unused imports that sneaked back in with the rebase
- Rebased on 4.x

Should be good to go now! 🤞

stof (approved):

GromNaN (approved):

GromNaN:
Thanks @Amoifr for working on this feature, this is much appreciated.

PR #579, opened 2026-04-29
Add Google Cloud formatter to default services

jannes-io:
| Q             | A
| ------------- | ---
| Branch?       | 4.x
| Bug fix?      | no
| New feature?  | yes
| Deprecations? | no
| Issues        | Fix #... 
| License       | MIT

Adds `GoogleCloudLoggingFormatter` to the default exposed services. The Google Cloud formatter is a class provided by monolog, so it makes sense to also expose it to the DI container so we can use it without having to add it to DI ourselves.

The Google Cloud formatter is pretty much essential for anyone running Symfony apps on Google Cloud infrastructure, otherwise the severity is unknown and the timestamp when the log is received is used instead of when the log was generated.

Before users had to manually register the `GoogleCloudLoggingFormatter`:
```php
    $services->set('app.logger.formatter.google', GoogleCloudLoggingFormatter::class);
```

And then use it in their monolog bundle config:
```
monolog:
    handlers:
        main:
            type: stream
            path: php://stderr
            level: info
            formatter: app.logger.formatter.google
```

After this change, users will no longer need to configure the `GoogleCloudLoggingFormatter` themselves and can just use `monolog.formatter.google` instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 
+* Add `GoogleCloudLoggingFormatter` as `monolog.formatter.google` to services
+
 ## 4.0.2 (2026-04-02)
 
 * Fix `TaggedIteratorArgument` deprecation warning when using `symfony/dependency-injection` 8.1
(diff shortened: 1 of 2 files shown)


GromNaN (approved):

HypeMC (approved):

GromNaN:
Thank you @jannes-io.

PR #583, opened 2026-06-13
Add a `timezone` config option

### LINKED ISSUE #419: Allow setting custom timezone in the configuration
The Monolog class have an setTimeZone method to adjust the time zone. that is currently not configureable using config files

Amoifr:
This adds a `timezone` config option to set the timezone used for the
timestamp of log records, exposing Monolog's `Logger` timezone support.

Fix #419


src/DependencyInjection/Configuration.php
```diff
@@ -357,6 +357,10 @@ public function getConfigTreeBuilder(): TreeBuilder
             ->fixXmlConfig('handler')
             ->children()
                 ->scalarNode('use_microseconds')->defaultTrue()->end()
+                ->scalarNode('timezone')
+                    ->info('The timezone used for the timestamp of every log record (e.g. "Europe/Paris"). Defaults to the PHP default timezone.')
```
GromNaN: "UTC" might be the preferred value.

```suggestion
                    ->info('The timezone used for the timestamp of every log record (e.g. "UTC" or "Europe/Paris"). Defaults to the PHP default timezone.')
```

src/DependencyInjection/Configuration.php
```diff
@@ -357,6 +357,10 @@ public function getConfigTreeBuilder(): TreeBuilder
             ->fixXmlConfig('handler')
             ->children()
                 ->scalarNode('use_microseconds')->defaultTrue()->end()
+                ->scalarNode('timezone')
```
GromNaN: `stringNode` was added in Symfony 7.3. It is more specific and improves the array shape accuracy.

```suggestion
                ->stringNode('timezone')
```

Amoifr:
Thanks for the review @GromNaN! Both applied — switched to `stringNode` (nice, didn't know about it) and added "UTC" to the example. 🙏

src/DependencyInjection/MonologExtension.php
```diff
@@ -58,6 +58,13 @@ public function load(array $configs, ContainerBuilder $container): void
 
             $container->setParameter('monolog.use_microseconds', $config['use_microseconds']);
 
+            if (null !== $config['timezone']) {
+                $prototype = $container->getDefinition('monolog.logger_prototype');
+                $prototype->setArgument(1, []);
+                $prototype->setArgument(2, []);
```
GromNaN, line 64: This defaults could be set in the config file, but it is less risky to add it only once the new option has been defined.
https://github.com/symfony/monolog-bundle/blob/3de7470e88d96dac71ad029fb65216da56edbb07/config/monolog.php#L43-L45

GromNaN (approved):

GromNaN:
Thank you @Amoifr.

